PR #1, opened 2021-02-12
Correct Registration/Sign-up Flow documentation

Jcambass:
If we're using the token to activate the user we need to use the `Auth.activate_user_from_token` function.


Jcambass:
Looks like CI tries to push a docker image for PRs. Might be something that you want to disable for non `main` branches :)

wmnnd:
Good catch, thanks for the PR!

Issue #3, opened 2021-02-12
Feature Request: Mailgun Support


mazz:
Could you add support for mailgun.org?

PR #5, opened 2021-02-12
Senders, Support Mailgun

Jcambass:
This patch adds support for Mailgun.
Mailgun is supported out of the box by Swoosh (https://github.com/swoosh/swoosh#adapters).

I understand if we don't want to support all adapters supported by swoosh.
If that is the eventual goal we might want to invest some time to make it easier to support a larger amount of adapters by drying up the custom logic around supporting different adapters in the UI and backend.

Closes https://github.com/pentacent/keila/issues/3


wmnnd:
Looking great, thank you!

PR #6, opened 2021-02-12
CI, Only Build and Push Container on Main

Jcambass:
This patch adds a conditional to the `Build and push Docker image` job so that this only runs on the `main` branch.
The indentation of the top-level `on:` option has also been fixed.

**Potential Followups**

We might want to look into generating docker version tags dynamically since currently, we push all merged changes to the same `0.1` tag.
If we introduce dynamically generated versions we technically could even use that to tag PRs if we really want that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: CI
 on:
   pull_request:
   push:
-      branches:
+    branches:
       - main
 
 jobs:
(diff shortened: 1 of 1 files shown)


wmnnd:
Awesome, thank you! 🙏

I also like your other suggestion. Right now, we’re still in pre-0.1 territory, once we reach 0.1, I would like to start building a public docker image that is only built from tagged releases.

PR #7, opened 2021-02-14
Refactor, Generalize SenderAdapter Code

Jcambass:
This patch moves sender adapter specific code into a centralized place so that adding support for more sender adapters becomes easier.
This should aid us in adding support for all adapters supported by Swoosh.
The UI currently still needs to define views and "translations" for the adapter names explicitly to allow more freedom.

One side effect to this is that we can't reference `Keila.Mailings.Sender.Config` in the adapter code since that would introduce cyclic dependencies.

Currently, the UI code needs to define some methods to handle the test sender adapter. That's not great but I wonder if we even need to support a test specific adapter or if we can use the `SMTP` adapter in all cases?

This is meant to provide a starting point for a discussion so I'm not saying that this needs to be the final solution.

Let me know what you think.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -152,7 +152,8 @@ if config_env() == :prod do
 
   # Deployment
   config :keila,
-    registration_disabled: System.get_env("DISABLE_REGISTRATION") not in [nil, 0, "", "false", "FALSE"]
+    registration_disabled:
+      System.get_env("DISABLE_REGISTRATION") not in [nil, 0, "", "false", "FALSE"]
 end
 
 if config_env() == :test do
--- a/lib/keila/mailings/sender_adapters/adapter.ex
+++ b/lib/keila/mailings/sender_adapters/adapter.ex
@@ -1,0 +1,76 @@
+defmodule Keila.Mailings.SenderAdapters.Adapter do
+  @moduledoc """
+  Defines a sender adapter.
+
+  Sender adapters are used by Keila to support sending campaings through various services.
+
+  `schema_fields/0` define configuration fields which will dynamically be added to the `Keila.Mailing.Sender.Config` schema.
+
+  `changeset/2` will be used to build a changeset for the configuration fields before storing them in a `Keila.Mailing.Sender.Config` schema.
+
+  `to_swoosh_config/1` allows to retrieve configuration fields from the `Keila.Mailings.Sender.Config` and build a config that will be passed to `Swoosh`.
+
+  ## Example
+      defmodule Sendgrid do
+        use Keila.Mailings.SenderAdapters.Adapter
+
+        @impl true
+        def name, do: "sendgrid"
+
+        @impl true
+        def schema_fields do
+          [
+            sendgrid_api_key: :string
+          ]
+        end
+
+        @impl true
+        def changeset(changeset, params) do
+          changeset
+          |> cast(params, [:sendgrid_api_key])
+          |> validate_required([:sendgrid_api_key])
+        end
+
+        @impl true
+        def to_swoosh_config(struct) do
+          [
+            adapter: Swoosh.Adapters.Sendgrid,
+            api_key: struct.sendgrid_api_key
+          ]
+        end
+      end
+
+  ## Configuration
+      config :keila, Keila.Mailings.SenderAdapters, adapters: [NewSenderAdapter]
+  """
+
+  defmacro __using__(_options) do
+    quote do
+      @behaviour unquote(__MODULE__)
+      import Ecto.Changeset
+      alias unquote(__MODULE__)
+    end
+  end
+
+  @type t :: module
+
+  @doc """
+  Returns the name of the sender adapter.
+  """
+  @callback name() :: String.t()
+
+  @doc """
+  Returns a list of exto schema fields that the adapter requires.
+  """
+  @callback schema_fields() :: keyword(atom())
+
+  @doc """
+  Builds a changeset for the sender adapter configuration.
+  """
+  @callback changeset(Ecto.Changeset.t(), %{optional(String.t()) => term()} | %{optional(atom()) => term()}) :: Ecto.Changeset.t()
+
+  @doc """
+  Builds a swoosh config from the passed sender adapter configuration.
+  """
+  @callback to_swoosh_config(struct()) :: keyword()
+end
--- a/lib/keila/mailings/sender_adapters/adapter.ex
+++ b/lib/keila/mailings/sender_adapters/adapter.ex
@@ -4,7 +4,7 @@ defmodule Keila.Mailings.SenderAdapters.Adapter do
 
   Sender adapters are used by Keila to support sending campaings through various services.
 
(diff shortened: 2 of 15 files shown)


config/runtime.exs
```diff
@@ -152,7 +152,8 @@ if config_env() == :prod do
 
   # Deployment
   config :keila,
-    registration_disabled: System.get_env("DISABLE_REGISTRATION") not in [nil, 0, "", "false", "FALSE"]
```
Jcambass, line 155: Some diffs like this come from `mix format`.

wmnnd:
I took a quick look yesterday and it looks like a very clever solution! 👏👏👏

I’ll review the merge request later today and give you some more detailed comments.

lib/keila/mailings/sender_adapters/schema.ex
```diff
@@ -0,0 +1,18 @@
+defmodule Keila.Mailings.SenderAdapters.Schema do
```
wmnnd, line 1: I’m not quite sure if this module is actually necessary. It seems like it is `use`d only once, so you might as well include `sender_adapters_fields` directly in `Keila.Mailings.Sender.Config` to avoid unnecessary complexity.

This should work directly in the schema definition:

```elixir
SenderAdapters.schema_fields() |> Enum.each(fn {name, type} -> field(name, type) end)
```
```

**Review by wmnnd — CHANGES_REQUESTED** [review]
Thank you for this very nice pull request!

It’s very well designed and I’ve added only a handful of comments :+1:

File: lib/keila_web/views/sender_view.ex [hunk 3]
```diff
@@ -1,3 +1,37 @@
...
+
+  def render_sender_adapter_form(form, "sendgrid") do
+    render("_sendgrid_config.html", form: form)
+  end
+
+  def render_sender_adapter_form(form, "mailgun") do
+    render("_mailgun_config.html", form: form)
+  end
+
+  # TODO: Replace me. Currently needed to support test sender adapter.
```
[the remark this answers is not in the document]
Jcambass: I'm not 100% sure.

I don't like having `if Mix.env` in the production code.

Since we don't want to leak KeilaWeb concerns into the sender adapters we have the following options:

- Add a module somewhere in KeilaWeb for each sender adapter and let that module have a `sender_adapter` method that returns the sender adapter. Then use KeilaWeb to configure a list of those modules instead of a list of sender adapters.

- Remove the need for the `test` sender adapter.

- Keep this as it is.

Now, for a hex library the furst approach would make the most sense so that it is easy to extend the lib.
But here we might just get away with keeping it as it is. We can still change that later if we need to.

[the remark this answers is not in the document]
wmnnd: How about adding a configuration `:keila, :env, :prod/:dev/:test`? Then this kind of code would be independent from how the environment is actually configured.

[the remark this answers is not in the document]
Jcambass: I think that's just hiding the real problem. It's basically code that is only there to support testing.

I think it's ok to keep it like that as long as it stays minimal. We can make the statement that the backend part of sender adapters is modular while the frontend currently isn't. The fact that we can at least guard with `Mix.env` is actually good since that at least means that we don't compile test-only code when running in production.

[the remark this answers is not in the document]
wmnnd: Okay, sounds good to me :+1:

Jcambass:
I think I addressed all the comments. A quick review of the documentation would be nice :)

lib/keila/mailings/sender_adapters/adapter.ex
```diff
@@ -0,0 +1,76 @@
...
+
+  @type t :: module
+
+  @doc """
+  Returns the name of the sender adapter.
+  """
+  @callback name() :: String.t()
+
+  @doc """
+  Returns a list of exto schema fields that the adapter requires.
```
wmnnd, line 63:
```suggestion
  Returns a list of Ecto schema fields required by the adapter.
```

lib/keila/mailings/sender_adapters/adapter.ex
```diff
@@ -0,0 +1,76 @@
+defmodule Keila.Mailings.SenderAdapters.Adapter do
+  @moduledoc """
+  Defines a sender adapter.
+
+  Sender adapters are used by Keila to support sending campaings through various services.
+
+  `schema_fields/0` define configuration fields which will dynamically be added to the `Keila.Mailing.Sender.Config` schema.
```
wmnnd, line 7:
```suggestion
  `schema_fields/0` defines configuration fields which will dynamically be added to the `Keila.Mailings.Sender.Config` schema.
```

lib/keila/mailings/sender_adapters/adapter.ex
```diff
@@ -0,0 +1,76 @@
...
+
+  `changeset/2` will be used to build a changeset for the configuration fields before storing them in a `Keila.Mailing.Sender.Config` schema.
+
+  `to_swoosh_config/1` allows to retrieve configuration fields from the `Keila.Mailings.Sender.Config` and build a config that will be passed to `Swoosh`.
```
wmnnd, line 11: I know I’m being super nitpicky now, but ideally the lines shouldn’t be much longer than 80 characters 😅

lib/keila/mailings/sender_adapters/adapter.ex
```diff
@@ -0,0 +1,76 @@
+defmodule Keila.Mailings.SenderAdapters.Adapter do
+  @moduledoc """
+  Defines a sender adapter.
+
+  Sender adapters are used by Keila to support sending campaings through various services.
+
+  `schema_fields/0` define configuration fields which will dynamically be added to the `Keila.Mailing.Sender.Config` schema.
+
+  `changeset/2` will be used to build a changeset for the configuration fields before storing them in a `Keila.Mailing.Sender.Config` schema.
```
wmnnd, line 9:
```suggestion
  `changeset/2` will be used to build a changeset for the configuration fields
  before storing them in a `Keila.Mailing.Sender.Config` schema.
```

lib/keila/mailings/sender_adapters/adapter.ex
wmnnd, line 11:
```suggestion
  `to_swoosh_config/1` retrieves configuration fields from
   `Keila.Mailings.Sender.Config` and builds the config that is passed to `Swoosh`.
```

wmnnd (commented):
Thank you for implementing the suggested changes. With this new small batch of suggestions we should be good to go :rocket:

lib/keila/mailings/sender_adapters/adapter.ex
Jcambass: No problem! I'm also im the 80 chars camp but I think I don't have it properly setup for elixir ind my editor.

Jcambass:
@wmnnd I don't have access to my computer st the moment so feel free to `mix format` and squash.

wmnnd (approved):

PR #8, opened 2021-02-15
Add Templates context, refacor HTML handling, add CSS parser

wmnnd:


lib/keila/mailings/builder.ex
```diff
@@ -117,22 +118,58 @@ defmodule Keila.Mailings.Builder do
...
-        {:ok, Solid.render(template, assigns) |> to_string()}
+        render_liquid(template, assigns)
       else
-        {:error, error = %Solid.TemplateError{}} ->
-          {:error, error.message}
+        {:error, error = %Solid.TemplateError{}} -> {:error, error.message}
       end
     rescue
-      _e -> {:error, "Unexpected Error"}
+      _e -> {:error, "Unexpected parsing prror"}
```
Jcambass, line 161:
```suggestion
      _e -> {:error, "Unexpected parsing error"}
```

wmnnd: Good catch, thanks!

Issue #15, opened 2021-02-20
Automatically run migrations + setup admin user on first launch


wmnnd:
Many applications such as Postgres automatically create an admin user when the env variables `ADMIN_USER` and `ADMIN_PASSWORD` are set on first launch.

Keila should automatically run migrations (via `Keila.ReleaseTasks.init()`) and create an admin user when the env variables are set.

wmnnd:
Implemented via #18

PR #18, opened 2021-02-22
Automatically run migrations + create root user from env

wmnnd:
Implements #15

### DIFF
--- a/priv/repo/seeds.exs
+++ b/priv/repo/seeds.exs
@@ -12,30 +12,45 @@ if Keila.Repo.all(Auth.Group) == [] do
   permission = Keila.Repo.insert!(%Auth.Permission{name: "administer_keila"})
   Keila.Repo.insert!(%Auth.RolePermission{role_id: role.id, permission_id: permission.id})
 
-  email = System.get_env("KEILA_USER")
-  password = System.get_env("KEILA_PASSWORD")
+  email =
+    case System.get_env("KEILA_USER") do
+      email when email not in ["", nil] ->
+        email
 
-  if email not in ["", nil] and password not in ["", nil] do
-    case Keila.Auth.create_user(%{email: email, password: password}) do
-      {:ok, user} ->
-        Keila.Auth.activate_user(user.id)
-        Keila.Auth.add_user_group_role(user.id, group.id, role.id)
-        Logger.info("Created root user with #{email}")
-
-      {:error, changeset} ->
-        Keila.ReleaseTasks.rollback(0)
-
-        errors =
-          Ecto.Changeset.traverse_errors(changeset, fn {msg, opts} ->
-            Enum.reduce(opts, msg, fn {key, value}, acc ->
-              String.replace(acc, "%{#{key}}", to_string(value))
-            end)
-          end)
-
-        Logger.error("Failed to create root user: #{inspect(errors)}")
-        Logger.flush()
-        System.halt(1)
+      _empty ->
+        Logger.warn("KEILA_USER not set. Creating root user with email: root@localhost")
     end
+
+  password =
+    case System.get_env("KEILA_PASSWORD") do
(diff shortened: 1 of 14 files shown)


Jcambass (approved):

priv/repo/seeds.exs
```diff
@@ -7,7 +7,36 @@ require Logger
 alias Keila.{Repo, Auth}
 
 if Keila.Repo.all(Auth.Group) == [] do
-  Keila.Repo.insert!(%Auth.Group{name: "root"})
+  group = Keila.Repo.insert!(%Auth.Group{name: "root"})
+  role = Keila.Repo.insert!(%Auth.Role{name: "root"})
+  permission = Keila.Repo.insert!(%Auth.Permission{name: "administer_keila"})
+  Keila.Repo.insert!(%Auth.RolePermission{role_id: role.id, permission_id: permission.id})
+
+  email = System.get_env("KEILA_USER")
```
pascalwengerter, line 15: These two are *nil*, perhaps you'd like to make them explicit / check if they're nil and use other values? (I've replaced them with a fake email & password for testing&development locally 😉)

pascalwengerter: Ok, GitHub's UI didn't do what I expected it to do. We're talking 

`email = System.get_env("KEILA_USER")`
and
`password = System.get_env("KEILA_PASSWORD")`

wmnnd: I’m not entirely sure I understand what you mean. In line 18, there is a check if these values are null and it doesn’t create a user in that case.

Or are you suggesting to create a default user in that case and print out a randomly generated password instead?

pascalwengerter: Ah sorry, yeah I wasn't very clear. I think it'd be great to create a default user with basic credentials if `System.get_env(...)` is nil.

This saves people from overwriting the variables and recreating the database/re-running the seeds (as I had to, eventually)

wmnnd: Sounds like a good idea. Right now, automatic migrations are disabled for `dev` mode but maybe even there we could enable it.

I’m also curious what you think, @Jcambass.

pascalwengerter: Just managed to get Keila set up and running via `docker-compose`, now I'm badly missing seeds 😄 perhaps I could create an user via `iex`? Let's add documentation for this, too!

Jcambass: @wmnnd I think there are a few important things to consider from an ops/user perspective. 
- Don't provide standard passwords.
- Documentation is great but a self-documenting setup process is better for most non-enterprise users.

Therefore I'm ok with providing a default username and a generated password upon running the seeds.

Personally, I think it might make sense to work towards being able to just start up the application and be able to perform the setup inside the application. 

This would probably be limited to domain-specific things like root username, password, and the system SMTP mailer setup. 
Things like DB connections and secret key bases would have a sane default/be generated upon the first start if possible. Those things are more configuration and depend a bit on the deployment scenario we choose to support while the more domain-related configurations are the same across all deployment environments.

wmnnd: @Jcambass Full ack on self-documenting processes. I don’t think I’m a fan of configuration-less setups that you would see on PHP CMS systems, for example. I don’t think it’s a good idea to have a moment where you have installed the application and anyone could just happen to visit your URL and access the configuration interface.

wmnnd: Added via 756971a1b128e441089c182c6c098bbcde8e39fc (defaulting to `root@localhost` and random password)

Jcambass: @wmnnd That's a valid point. I think the application which I used and had that pattern were onces which were by design only on the local network which makes it a bit less of a problem. 

In the end both approaches have pros and cons. So I understand if we don't want to take this path.

PR #20, opened 2021-02-22
Concurrency, Prevent double sending of Campaigns

Jcambass:
While trying to implement the deliver campaign later feature I noticed that we didn't properly guard against double sendings of campaigns.
The idea here is to use lock on the campaign and use `sent_at` as a guard to ensure that the campaign is not sended twice.

This came up since I tried to implement the send later feature for campaigns based on a new `deliver_at` field in combination with a Oban cron job rather than just scheduling jobs for each scheduled campaign.
The benefit of that is more control over sending of those scheduled campaigns and having the knowledge about the time the campaign will be delivered inside the domain rather than having it only inside Oban (or introducing a potential disconnect between those sources of truth).
Based on my experience this pattern is better scalable and adjustable for different needs than just scheduling jobs in the future.

### DIFF
--- a/lib/keila/mailings/mailings.ex
+++ b/lib/keila/mailings/mailings.ex
@@ -170,54 +170,55 @@ defmodule Keila.Mailings do
   """
   @spec deliver_campaign(Campaign.id()) :: {:error, :no_recipients} | {:error, term()} | :ok
   def deliver_campaign(id) when is_id(id) do
-    {:ok, campaign} =
-      id
-      |> get_campaign()
-      |> change(sent_at: DateTime.truncate(DateTime.utc_now(), :second))
-      |> Repo.update()
+    result =
+      Repo.transaction(fn ->
+        case get_and_lock_campaign(id) do
+          %Campaign{sent_at: nil} = campaign -> do_deliver_campaign(campaign)
+          %Campaign{} -> Repo.rollback(:already_sent)
+        end
+      end)
 
-    stream = Keila.Contacts.stream_project_contacts(campaign.project_id, [])
-
-    case do_deliver_campaign(stream, campaign) do
-      {:ok, :ok} ->
-        :ok
-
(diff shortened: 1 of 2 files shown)


lib/keila/mailings/mailings.ex
```diff
@@ -64,6 +64,15 @@ defmodule Keila.Mailings do
     Repo.get(Campaign, id)
   end
 
+  @doc """
+  Retrieves and locks the Campaign with the given `id`.
+  """
+  @spec get_and_lock_campaign(Campaign.id()) :: Campaign.t() | nil
+  def get_and_lock_campaign(id) when is_id(id) do
```
wmnnd, line 71: Should this really be an exported function? I think `defp` might suffice here.

lib/keila/mailings/mailings.ex
```diff
@@ -170,54 +179,50 @@ defmodule Keila.Mailings do
...
+    result =
+      Repo.transaction(fn ->
+        case get_and_lock_campaign(id) do
+          %Campaign{sent_at: nil} = campaign -> do_deliver_campaign(campaign)
+          %Campaign{} -> Repo.rollback(:already_sent)
+        end
+      end)
 
-        {:error, reason}
+    case result do
```
wmnnd, line 190: I like this pattern dealing with the odd `{:ok, :ok}` from transactions :+1:

wmnnd (commented):
Looking pretty good! Does the `:already_sent` status need some kind of representation in the UI, though?

Jcambass:
If I understand it correctly currently the UI sends the campaigns async without really handling errors or similar. So I think that's ok for the moment since it would just redirect to the `stats` page of the already sent campaign if there are two users sending the campaign at the same time.

wmnnd (approved):

wmnnd:
Thank you, excellent work again!

PR #21, opened 2021-02-22
Add note about Postgres to README

pascalwengerter:
This one caused me quite some (avoidable) trouble since it took me a while to figure out why `mix.setup` would fail - turns out I didn't have Postgres installed yet!


wmnnd:
Thank you :smile:

PR #30, opened 2021-02-25
Add purge config for TailwindCSS to reduce CSS bundle size

### LINKED ISSUE #28: Introduce PurgeCSS for Tailwind
When checking if Webpack works on #26 I realized that the `../css/app.ass/` is ~4MB in size. This is due to Tailwind creating a lot of helper classes and can be optimized for production use via PurgeCSS.

Here's a [guide](https://pragmaticstudio.com/tutorials/adding-tailwind-css-to-phoenix), happy to take care of this eventually (unless someone else is quicker than me 🤓 )

pascalwengerter:
@wmnnd this should fix #28 

Could you check it out and run it locally using `MIX_ENV=prod mix phx.server` and confirm that it changes from something around this

Screenshot 2021-02-25 at 20 00 42

to something looking more like this?

Screenshot 2021-02-25 at 20 00 34

I did have some issues with what I suspect was my local webpack cache, so ping me if it doesn't (and I'll share more insights)!

### DIFF
--- a/assets/package.json
+++ b/assets/package.json
@@ -3,7 +3,7 @@
   "description": " ",
   "license": "MIT",
   "scripts": {
-    "deploy": "webpack --mode production",
+    "deploy": "NODE_ENV=production webpack --mode production",
     "watch": "webpack --mode development --watch"
   },
   "dependencies": {
--- a/assets/tailwind.config.js
+++ b/assets/tailwind.config.js
@@ -1,5 +1,10 @@
 module.exports = {
-  purge: [],
+  purge: [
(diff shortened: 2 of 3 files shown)


wmnnd:
Running `npm i` / `npm ci` on your branch keeps getting stuck at `reify:tailwindcss: timing reifyNode:node_modules/lodash Completed in 2928ms`

wmnnd:
Looking great! I ran it locally and also got an `app.css` with `26.5 KiB` :+1: 

The purge rules also look reasonable, so I’ll merge it now. Thank you! 🙏

PR #31, opened 2021-02-25
Fix 'email is invalid' error when running seeds

pascalwengerter:
Just ran into `[error] Failed to create root user: %{email: ["is invalid"]}` when trying to run the seeds on the `0.2` release.

This PR should fix the problem, didn't bother with opening a corresponding issue.

@wmnnd @Jcambass I really like this approach in terms of user ergonomics. It's a sweet mix of _not_ having an openly configurable application waiting to be abused (since there's a password & email set) and not having to know Elixir/jumping into `IEX` to manually create records! 🙌🏽

### DIFF
--- a/priv/repo/seeds.exs
+++ b/priv/repo/seeds.exs
@@ -18,7 +18,9 @@ if Keila.Repo.all(Auth.Group) == [] do
         email
 
       _empty ->
-        Logger.warn("KEILA_USER not set. Creating root user with email: root@localhost")
+        email = "root@localhost.dev"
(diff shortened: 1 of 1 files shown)


wmnnd:
Oops, this is what happens when you test something and *then* you add another line of code. I’ll tag this as 0.2.1 right away.

Jcambass:
One note here: I think we should not use `root@localhost.dev` since `.dev` is a real tld. I think `root@localhost.local` would be better.

wmnnd:
@Jcambass Agreed, I had already changed it back to root@localhost :-)

Issue #11, opened 2021-02-19
Send later feature


wmnnd:
Instead of sending campaigns immediately, there should be an option to automatically send campaigns at a later date.

This should be implemented as an Oban scheduled job.

- [x] Implement `Mailings.campaign_deliver_later` with Oban scheduled job for sending
- [x] Implement interface for scheduling campaigns from campaign editor. Include option to select timezone.

Jcambass:
@wmnnd What's your take on using scheduled jobs vs using a periodic CRON job to deliver mails?

I first wanted to go for a CRON job solution but I wonder if it is already worth it to go that route.
I collected the following pros/cons:

**Scheduled Job**

Pros:
- Can be unique to make sure we don't send the campaign twice (already handled somewhat but still nice to have).
- Concurrency is limited by the queue configuration.

Cons:
- Fills up the queue with jobs that potentially stay around for long.
- Knowledge that the campaign has been scheduled for delivery needs to be duplicated (the job and a column in the campaign).-

**CRON job**

Pros:
- One source of truth. We only need the `deliver_at` column in campaigns to know that a campaign has been scheduled and when it has to be sent.
- Only one job is inserted periodically to poll for campaigns that need to be delivered.
- We can finetune the poll interval for campaigns.

Cons:
- We need to handle max concurrency ourselves.
- We need to make a tradeoff between limiting the number of campaigns to deliver on each interval and the duration of the interval.

I originally was thinking that a periodic cron job would be better but I think it really depends on how time accurate we need to be and on how many scheduled campaigns we expect. What do you think?

wmnnd:
Thanks for the write-up!

I still like your suggestion of using a single source of truth (`deliver_at`) in combination with periodic checks whether a campaign should be sent. Oban supports CRON-style jobs, so this should also be pretty straightforward to implement. And an interval of once every minute or once every five minutes should not add too much overhead.
Since sending a campaign will always take a while if you have lots of recipients, I don’t think higher granularity would make sense.

At this stage, I’m also not super concerned about the concurrency implications.
I think eventually we’ll have to find a solution to make sure that campaigns are not being sent purely based on FIFO but instead implement something that allows multiple campaigns to be delivered at the same time (i.e. instead of delivering campaigns A and B as `AAAAAABBBBB`, deliver them as `ABABABABABA`). But this is beyond the scope of this issue and something we’ll have to implement regardless of the approach we choose.

[...]

PR #32, opened 2021-02-26
Keila Core, Send Campaign Later

Jcambass:
This patch adds the core foundation to allow scheduling campaigns to be sent at a later time.

The original idea was to schedule a job for each scheduled campaign.
This PR takes a different approach and implements a periodic worker that checks for scheduled campaigns and sends them.
The reasoning, as well as a non-comprehensive list of pros and cons, can be found in https://github.com/pentacent/keila/issues/11.

**Oban Upgrade**
I also upgraded Oban to the latest stable version since there have been some breaking changes on how queues and cron jobs are configured and a DB migration.
It's probably easier to upgrade now than later when there is more adoption so I snaked this one in here.
See: https://hexdocs.pm/oban/2.4.3/changelog.html#2-4-3-2021-02-07

**UI**
This patch intentionally does not implement a UI for this feature yet.


wmnnd (changes requested):
This time I really have only two very minor remarks for consistency/clarity:

1) I found the name `cluster` for the scheduling queue a little confusing.
2) The property for sent/delivered campaigns is `sent_at`, so `send_at` would make more sense than `deliver_at`. But in order to avoid any kind of confusion, maybe `scheduled_for` would be even nicer.

Other than that, once again a very nice PR!
One thing that I really enjoy about working with Elixir and Phoenix is exemplified here: You can build your application like you would build a library, the web interface is really a UI layer on top of that :smiley:

Jcambass:
1. The idea behind `cluster` here is to have a queue where non critical jobs can be performed. As you wrote in the issue: it's not a massive problem if we're a bit late with sending a scheduled campaign and keeping `mailer` going is more important. Instead of adding one queue for every feature we add a shared queue like `cluster` can be used.

That bein said this idea mostly makes sense for things like Sidekiq where each worker process listens to a specific queue. With Oban that's not the case AFAIK and having many queues is not costly anymore (https://hexdocs.pm/oban/changelog.html#2-4-0-2021-01-26).
So if you can think of a good nane for this queue (and maybe size) we can change it.

wmnnd:
Thanks! Maybe something plain like `scheduled_campaigns` would work? And we might actually try to go with a max concurrency of 1 for that, since they don’t really need to run in parallel, do they?

Jcambass:
@wmnnd Do you want to go for that style of almost one queue per worker or should we add a queue for all CRON jobs like `periodic` we can start with 1 concurrency and increase it if we ever need more periodic worker (which ideally are always fast and just enqueue/task off other work).

wmnnd:
How about we name it simply `cron` then, with concurrency 1, and if we ever find that’s not suitable, we can always change it to something more specific again :-)

Jcambass:
@wmnnd Sure `cron` is fine as well, personally I opted for `periodic` over `cron` to not commit a wrong idea that the `queue` name of workers configured in the `cron` plugin needs to be `cron` as well. But if you favour it over `periodic` we can change it.

In regards to renaming the queue this would probably need to be done by adding a new queue in one release and then removing it in a second release (or providing enough instructions for users to do that) in order to not break other users installations, so if possible I wanted to choose a good name now :)

wmnnd:
I think you’re right and `periodic` might also avoid confusion between the queue and the plugin :+1:

Jcambass:
@wmnnd Changed to `periodic` again :)

wmnnd (approved):

wmnnd:
Thank you once again for this PR! Do you also want to add the UI portion of this feature? Otherwise I’ll take care of it this week :smiley:

Jcambass:
@wmnnd I have something started but it's not much yet and I don't think that I will be able to get to it this week so feel free to take it.

PR #34, opened 2021-03-07
Feature/send later UI

wmnnd:
Here’s the preliminary interface for scheduling campaigns:
Screenshot_2021-03-07 Open Source Newsletter Tool · Keila

There’s no code in this MR yet (the commits are just here because it’s based on the WYSIWYG branch), but I’m looking for feedback on a little API question. Right now, (un)scheduling a campaign works via `schedule_campaign/2`. I would like to enforce rules for scheduling campaigns, e.g. that they can only be scheduled for times that are at least `n` minutes in the future or that they can’t be rescheduled/unscheduled if they’ve already been scheduled to be sent out in less than `n` minutes.

My idea was to add an option to `schedule_campaign/2` (turning it into `schedule_campaign/3`) that would allow us to pass this `n` and then use it in a changeset validation.
I don’t think I’ve seen this pattern before, so I’m curious what you think.

An obvious alternative would be to use the application configuration for setting `n`.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -26,7 +26,7 @@ config :keila, Keila.Id,
 
 config :keila, Keila.Mailings,
   # Minimum offset in seconds between current time and allowed scheduling time
-  campaign_schedule_offset: 300
+  min_campaign_schedule_offset: 300
 
 config :keila, Keila.Mailings.SenderAdapters,
   adapters: [
--- a/config/config.exs
+++ b/config/config.exs
@@ -64,6 +64,9 @@ config :keila, Oban,
      ]}
   ]
 
+# Use Timezone database
+config :elixir, :time_zone_database, Tzdata.TimeZoneDatabase
+
 # Import environment specific config. This must remain at the bottom
 # of this file so it overrides the configuration defined above.
 import_config "#{config_env()}.exs"
--- a/config/config.exs
+++ b/config/config.exs
@@ -24,6 +24,10 @@ config :keila, Keila.Id,
   salt: "bF4QzDjqV",
   min_len: 8
 
+config :keila, Keila.Mailings,
+  # Minimum offset in seconds between current time and allowed scheduling time
+  campaign_schedule_offset: 300
+
 config :keila, Keila.Mailings.SenderAdapters,
   adapters: [
     Keila.Mailings.SenderAdapters.SMTP,
--- a/config/test.exs
+++ b/config/test.exs
@@ -33,6 +33,9 @@ config :argon2_elixir, t_cost: 1, m_cost: 8
 # Disable Oban Queues
 config :keila, Oban, queues: false, plugins: false
 
+config :keila, Keila.Mailings,
+  campaign_schedule_offset: -10
+
 # Only use test and smtp Sender Adapters
 config :keila, Keila.Mailings.SenderAdapters,
   adapters: [
--- a/config/test.exs
+++ b/config/test.exs
@@ -33,8 +33,8 @@ config :argon2_elixir, t_cost: 1, m_cost: 8
 # Disable Oban Queues
 config :keila, Oban, queues: false, plugins: false
 
-config :keila, Keila.Mailings,
-  campaign_schedule_offset: -10
+# Allow scheduling campaigns at utc_now
+config :keila, Keila.Mailings, min_campaign_schedule_offset: -10
 
 # Only use test and smtp Sender Adapters
 config :keila, Keila.Mailings.SenderAdapters,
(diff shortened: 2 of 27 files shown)


Jcambass:
I think the concept sounds good. I would lean towards an application config for this. I think it's a value that should stay consistent actoss all call sites and therefore be configured once.

wmnnd:
As you suggested, @Jcambass, I made the threshold a global configuration option instead of a parameter.

This PR now also includes some neat datetime improvements, allowing client-side handling of datetime rendering. Here is a screenshot of the slightly revamped campaign index page from my browser with an en-US locale:
Screenshot_2021-03-10 Campaigns · Keila

config/config.exs
```diff
@@ -24,6 +24,10 @@ config :keila, Keila.Id,
   salt: "bF4QzDjqV",
   min_len: 8
 
+config :keila, Keila.Mailings,
+  # Minimum offset in seconds between current time and allowed scheduling time
+  campaign_schedule_offset: 300
```
Jcambass, line 29: I wonder if there would be a more descriptive name we could use but I honestly can't reallt come up with anything better on the spott.

config/test.exs
```diff
@@ -33,6 +33,8 @@ config :argon2_elixir, t_cost: 1, m_cost: 8
 # Disable Oban Queues
 config :keila, Oban, queues: false, plugins: false
 
+config :keila, Keila.Mailings, campaign_schedule_offset: -10
```
Jcambass, line 36: Why do we configure a negative offset here? I think this would not be allowed in real usage so if we do it here to facilitate a single test or something we should at least put a comment here.

Ideally we would only configure this for the one test but I guess since it's a global config it's a bit harder (unless we would built in escape hatches that allow us to pass the param as an optional config but I'm not a hughe fan of that either).

lib/keila/mailings/schemas/campaign.ex
```diff
@@ -47,6 +47,45 @@ defmodule Keila.Mailings.Campaign do
...
+
+    case DateTime.compare(threshold, scheduled_for) do
+      :gt -> [scheduled_for: "must be at least #{offset} seconds in the future"]
+      _ -> []
+    end
+  end
+
+  defp campaign_schedule_offset_and_threshold() do
+    offset =
+      Application.get_env(:keila, Keila.Mailings, []) |> Keyword.get(:campaign_schedule_offset, 0)
```
Jcambass, line 85: Is `0` a sane default here? IMHO think we don't really want to fallback to anything here since we probably want to always have a offset unless we specify otherwise.

config/config.exs
wmnnd: I went through three iterations or so to come up with the name, but it’s really hard to describe the concept - so I’m open for better suggestions :smile: 
Maybe `min_campaign_schedule_offset` would be a bit clearer?

config/test.exs
wmnnd: I’d say that the offset is mostly a matter of convenience and avoiding user confusion rather than ensuring system integrity. In the same vein, I think wanting to schedule tests at `utc_now` is very reasonable in the testing environment; having an offset of `0` might make tests fail occasionally - hence the negative value.

lib/keila/mailings/schemas/campaign.ex
wmnnd: Mhm … good question. Maybe a default is actually not even necessary at all since the five-minute default is set in `config.exs`. What do you think?

config/test.exs
Jcambass: Good points. What do you think about adding a comment here explaining this?

lib/keila/mailings/schemas/campaign.ex
Jcambass: I think we should rather raise to surface any configuration errors.

config/config.exs
Jcambass: Yeah `min_campaign_schedule_offset` sounds like good enough 👍🏻

lib/keila/mailings/schemas/campaign.ex
wmnnd: Using Keyword.fetch! now to raise errors.

config/test.exs
wmnnd: Good idea, comment added!

config/config.exs
wmnnd: :partying_face:

Jcambass (approved):
Great work! 🥳

Issue #48, opened 2021-06-14
Make system email configurable


wmnnd:
Right now, `hello@keila.io` is used in some instances for sending system emails. This should be configurable via an environment variable instead.

pascalwengerter:
Would you want to (re)use `MAILER_SMTP_USER` there or add a new, dedicated ENV var @wmnnd ?

wmnnd:
Good question! I think for now just using that would be sufficient :smiley:

wmnnd:
Implemented in #145.

PR #49, opened 2021-06-14
Read sender mail from ENV var

pascalwengerter:
Adresses #48 in a naive way

### DIFF
--- a/lib/keila/auth/emails.ex
+++ b/lib/keila/auth/emails.ex
@@ -3,6 +3,8 @@ defmodule Keila.Auth.Emails do
   # FIXME Don't depend on Web App here
   import KeilaWeb.Gettext
 
+  sender_mail = System.fetch_env!("MAILER_SMTP_USER")
+
   @spec send!(atom(), map()) :: term() | no_return()
   def send!(email, params) do
     email
@@ -13,7 +15,7 @@ defmodule Keila.Auth.Emails do
   @spec build(:activate, %{url: String.t(), user: Keila.Auth.User.t()}) :: term() | no_return()
   def build(:activate, %{user: user, url: url}) do
     new()
-    |> from({"Keila", "hello@keila.io"})
+    |> from({"Keila", @sender_mail})
     |> subject(dgettext("auth", "Please Verify Your Account"))
     |> to(user.email)
     |> text_body(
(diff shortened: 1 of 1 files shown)


wmnnd:
Hey, thanks for giving it a shot! The reason it can’t work like this is because of the way module attributes (`@something`) work. They are set at compile time - so at a point when the env variable is not yet available.

I suggest defining a function `sender_mail/1` in `Keila.Auth.Emails` and instead of using `System.fetch_env!/1` using `Application.get_env/2` to retrieve it fom the runtime config.

I would also suggest adding a new env variable, `MAILER_FROM_EMAIL` in `runtime.exs` and use that explicitly instead of automatically using the SMTP user name. You can find the relevant bit in the runtime config here: https://github.com/pentacent/keila/blob/03317ad783df7ff51f542ad5168f6a820eab046c/config/runtime.exs#L41
In addition, you’d have to add `config(:keila, Keila.Mailer, from_email: "hello@keila.io")` to the default `config.exs` so that it will continue to work in `dev` and `test` :smiley:

pascalwengerter:
> Hey, thanks for giving it a shot! The reason it can’t work like this is because of the way module attributes (`@something`) work. They are set at compile time - so at a point when the env variable is not yet available.
> 
> I suggest defining a function `sender_mail/1` in `Keila.Auth.Emails` and instead of using `System.fetch_env!/1` using `Application.get_env/2` to retrieve it fom the runtime config.
> 
> I would also suggest adding a new env variable, `MAILER_FROM_EMAIL` in `runtime.exs` and use that explicitly instead of automatically using the SMTP user name. You can find the relevant bit in the runtime config here:
> 
> https://github.com/pentacent/keila/blob/03317ad783df7ff51f542ad5168f6a820eab046c/config/runtime.exs#L41
> 
> In addition, you’d have to add `config(:keila, Keila.Mailer, from_email: "hello@keila.io")` to the default `config.exs` so that it will continue to work in `dev` and `test` 😃

Let me see if I can come up with an acceptable solution in the upcoming days(?) ;)

pascalwengerter:
@wmnnd feel free to take over here, the PR has been stale for too long and I won't be able to give it attention anytime soon :/ sry!

Issue #53, opened 2021-06-25
Add Click Tracking


wmnnd:
There should be a new Context, `Keila.Tracking`.

This context should have the following functions:

- [x] `register_link(campaign_id, url)`
- [x] `track_click(recipient_id, link_id)`
- [x] `track_open(recipient_id)`

wmnnd:
Implemented via #54

PR #54, opened 2021-06-28
Implement click/open tracking for emails

wmnnd:
Screenshot 2021-06-28 at 13-04-37 Open Source Newsletter Tool · Keila
Implements #53


wmnnd:
This PR introduces tracking and enables it for all campaigns. There should be an option for disabling tracking on a per-campaign and per-contact level. I’ll create additional issues for that.

PR #88, opened 2021-11-15
Switch to heex and improved UI

wmnnd:
This MR switches all templates to heex and introduces an improved UI with a sidebar and more consistent design.


wmnnd:
Here’s what the new design looks like :partying_face: 
Screen Shot 2021-11-15 at 21 04 23

Issue #75, opened 2021-10-01
Full API


wmnnd:
## Campaigns

`GET /api/v1/campaigns` - Fetch list of campaigns

`POST /api/v1/campaigns` - Create a new campaign
`GET /api/v1/campaigns/:id` - Fetch existing campaign
`PATCH /api/v1/campaigns/:id` - Update existing campaign

`POST /api/v1/campaigns/:id/actions/send` - Send existing campaign
`POST /api/v1/campaigns/:id/actions/schedule` - Schedule existing campaign

## Contacts
`GET /api/v1/contacts` - Fetch list of contacts – needs pagination and filtering

`POST /api/v1/contacts/actions/import` - Import contacts
`POST /api/v1/contacts/actions/export` - Export contacts

`POST /api/v1/contacts` - Create a new contact
`GET /api/v1/contacts/:id` - Fetch existing contact
`PATCH /api/v1/contacts/:id` - Update existing contact

wmnnd:
This PR now includes APIs for managing Contacts, Campaigns, and Segments and I think it’s a good starting point. I would be very happy about some feedback @panoramix360 & @gbottari  :smiley:

PR #93, opened 2021-12-04
Implement API

wmnnd:
Implementation of #75


wmnnd:
This PR now includes APIs for managing Contacts, Campaigns, and Segments and I think it’s a good starting point. I would be very happy about some feedback @panoramix360 & @gbottari :smiley:

panoramix360:
I'll try to look into it tomorrow :D

lib/keila_web/helpers/normalizer/normalizer.ex
```diff
@@ -0,0 +1,93 @@
...
+  import KeilaWeb.ApiNormalizer.SchemaMapper
+
+  def init(opts) do
+    opts
+  end
+
+  def call(conn, opts) do
+    opts
+    |> Keyword.get(:normalize, [])
+    |> Enum.reverse()
```
panoramix360, line 13: Is this really necessary? What is the purpose of reversing the normalize options?

lib/keila_web/helpers/normalizer/normalizer.ex
```diff
@@ -0,0 +1,93 @@
...
+    else
+      {:error, status: 400, title: "Invalid filter"}
+    end
+  end
+
+  def normalize(:contacts_filter, _) do
+    {:ok, :filter, %{}}
+  end
+
+  def normalize({:data, name}, %{"data" => data}) do
```
panoramix360, line 79: This normalize function is not checking and validating the fields of the model in question, maybe it's a good idea to do something like this and return "Invalid data" or something like that

lib/keila_web/views/api_view.ex
```diff
@@ -0,0 +1,125 @@
...
+    %{"data" => segment_data(segment)}
+  end
+
+  def render("errors.json", %{errors: errors}) do
+    %{
+      "errors" => Enum.map(errors, &error_object/1)
+    }
+  end
+
+  defp contact_data(contact) do
```
panoramix360, line 59: What do you think in separating this _data methods on the correspondent modules?

In this way we can tie together the map JSON to the contact schema.

lib/keila_web/controllers/api_controller.ex
```diff
@@ -0,0 +1,231 @@
...
+    conn
+    |> put_status(204)
+  end
+
+  #
+  # Campaign functions
+  #
+
+  plug ApiNormalizer,
+       [normalize: [{:data, :campaign}]]
```
panoramix360, line 70: I don't see the implementation of :campaign and :segment normalizer, is this correct?

lib/keila_web/controllers/api_controller.ex
```diff
@@ -0,0 +1,231 @@
...
+    |> put_status(204)
+  end
+
+  def deliver_campaign(conn, %{"id" => id}) do
+    # TODO immediate feedback on missing sender or insufficient credits
+    campaign = Mailings.get_project_campaign(project_id(conn), id)
+
+    if campaign do
+      Mailings.deliver_campaign_async(campaign.id)
+      put_status(conn, 204)
```
panoramix360, line 128: Maybe in this case return status code 202 to indicate that the request was accepted but will be processed.

lib/keila_web/controllers/api_controller.ex
```diff
@@ -0,0 +1,231 @@
...
+    else
+      send_404(conn)
+    end
+  end
+
+  def schedule_campaign(conn, params = %{"id" => id}) do
+    campaign = Mailings.get_project_campaign(project_id(conn), id)
+
+    if campaign do
+      scheduled_for = get_in(params, ["data", "scheduledFor"])
```
panoramix360, line 138: In this case, makes sense to add a normalizer for scheduledFor?

lib/keila_web/router.ex
```diff
@@ -151,11 +147,40 @@ defmodule KeilaWeb.Router do
     get "/c/:encoded_url/:recipient_id/:link_id/:hmac", TrackingController, :track_click
   end
 
-  scope "/api", KeilaWeb do
+  pipeline :api do
+    plug :accepts, ["json"]
+  end
+
+  scope "/api/v1", KeilaWeb do
```
panoramix360, line 154: Just an idea, what do you think of putting this version `v1` based on a configuration of the project, so when we increase the number of the version api, the url updates as well.

I'm not really sure if this is possible

panoramix360:
The purpose of this PR is to enable Keila to be integrated with other backends, right? Really nice :D

Some ideas or thoughts in general:
- I think that in the future, we can break `normalizer.ex` in more parts to separate the concern for each API, in this way we can define more normalizers for each route without mixing all the logic together.
- The same above applies to `api_controller.ex` in the future
- Maybe some method to inject a "data" property in all maps on `api_view.ex` can be a good idea to maintain the structure.
- Maybe adding some comments on `api_controller.ex` methods can be a good way to document the structure of the requests like the body and response.
- IDK if it's worth it, but adding a `delete_batch` to contacts to delete a list of contacts by id could be nice.

I wrote some comments on your code, but sorry for putting too many haha :)

Feel free to let me know what do you think \o

wmnnd:
Those are some very good points! I think I’ll try refactoring the code a bit, including special Schema modules and something to create OpenAPI schemas.

wmnnd:
Updated PR with OpenAPI will hopefully be done tomorrow :partying_face: 
Screenshot 2021-12-27 at 20-05-54 Swagger UI

wmnnd:
Thanks so much for your valuable feedback, @panoramix360! I have now done an extensive refactoring of the API controllers, making them much more modular, and adding Open API specs :partying_face:

panoramix360:
I'm glad that it helped :D

I'll look for your modifications on this week :tada:

Issue #87, opened 2021-11-12
Allow updating existing contacts on import


wmnnd:
It should be possible to update existing contacts on import, i.e. merging/replacing custom data or updating first/last name.

gbottari:
Hi, I'm interested in working on this issue if that's ok.

If I understand correctly, you need to check the database for existing contacts (by e-mail, I presume) and replace their infos (like first/last name). That logic would go into `Keila.Contacts.Import`, correct? Any tips would be welcome!

wmnnd:
Hey @gbottari, awesome, that would be very cool!

I think the import feature should by expanded with the following features:

1. Modify `Keila.contacts.Import.import_csv/2` to update existing entries with upserts (I think this would make sense as the default behavior)
2. Optionally: Add an option to `import_csv` that would allow for upserts to be disabled + a checkbox in the UI to reflect that
3. If you’re up for it, we should add a new field `external_id` (with a type adequate for storing arbitrary IDs, e.g. varchar(36)) with a unique index (project_id, external_id) that takes precedence over the email address when it comes to updating. This would allow people to keep Keila in sync with an external system and also allow for emails to be changed during imports.

Oh, and you should make sure to work off my branch with the updated UI if you want to make any UI changes. But I’ll probably merge it tomorrow :slightly_smiling_face:

gbottari:
Hey @wmnnd! Me and @panoramix360 are working on this issue now. We figured that we could use Postgres' own upsert feature through Ecto (https://hexdocs.pm/ecto/Ecto.Repo.html#c:insert/2-upserts). What we ended up with is the following:

`Repo.insert(changeset, returning: false, on_conflict: {:replace_all_except, [:email, :project_id]}, conflict_target: [:email, :project_id])`

We'll also look into the other tasks you mentioned.

PR #94, opened 2021-12-06
Feature/upsert on import

gbottari:
Implements the import feature described at #87.

- now on `Keila.Contacts.import_csv` there's an extra parameter called `:on_conflict` that can either be `:replace` or `:ignore`
- if set to `:replace`, duplicated contacts (as detected by the e-mail) will be replaced with the latest ones found in the import file
- if set to `:ignore`, duplicated contacts found will always be skipped
- on the UI there's now a checkbox that controls this behavior and is set by default to replace

### DIFF
--- a/lib/keila/contacts/import.ex
+++ b/lib/keila/contacts/import.ex
@@ -16,11 +16,12 @@ defmodule Keila.Contacts.Import do
           :ok | {:error, String.t()}
   def import_csv(project_id, filename, opts) do
     notify_pid = Keyword.get(opts, :notify, self())
+    on_conflicts = Keyword.get(opts, :on_conflict, :replace)
 
     Repo.transaction(
       fn ->
         try do
-          import_csv!(project_id, filename, notify_pid)
+          import_csv!(project_id, filename, notify_pid, on_conflicts)
         rescue
           e in NimbleCSV.ParseError ->
             Repo.rollback(e.message)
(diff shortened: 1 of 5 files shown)


lib/keila_web/live/contact_import_live.ex
```diff
@@ -19,11 +19,12 @@ defmodule KeilaWeb.ContactImportLive do
...
   @impl true
-  def handle_event("validate", _params, socket) do
-    {:noreply, socket}
+  def handle_event("validate", %{"import" => import_options}, socket) do
+    {:noreply, assign(socket, :import_replace, import_options["replace"] == "true")}
   end
 
-  def handle_event("import", _params, socket) do
+  def handle_event("import", %{"import" => import_options}, socket) do
+    # IO.inspect(params)
```
wmnnd, line 27: Leftover `IO.inspect` :smile:

lib/keila/contacts/import.ex
```diff
@@ -16,11 +16,12 @@ defmodule Keila.Contacts.Import do
           :ok | {:error, String.t()}
   def import_csv(project_id, filename, opts) do
```
wmnnd, line 17: Could you document the options here?

wmnnd (changes requested):
Looking great! Thanks for this PR. :+1: 

I just added two small comments, please let me know if you can take care of them :slightly_smiling_face:

lib/keila_web/live/contact_import_live.ex
panoramix360: Done!

lib/keila/contacts/import.ex
panoramix360: Done!

panoramix360:
@wmnnd if you have another issue in mind for us (me and @gbottari) let us known.

In the meantime, we'll search for another one.

wmnnd (approved):

wmnnd:
Thank you for contributing to Keila, @gbottari and  @panoramix360! :pray:

Issue #95, opened 2021-12-09
Add DevContainer support


panoramix360:
I noticed that the project has a Dockerfile and docker-compose.yml, and was wondering if you guys know DevContainer from Microsoft?

Is a [tool](https://code.visualstudio.com/docs/remote/containers) to Visual Code that lets you develop inside a container, it's pretty neat.

To build/develop the project on my machine when contributing in the last PR I configured a devcontainer.json on the project to run it inside a container, it uses docker-compose by default.

Do you think that it'll be nice to have this configuration on the project? I can open a PR for this.

Thank you! :smiley:

wmnnd:
I think that’s a neat idea to make developing Keila even easier! Feel free to create a PR, including a note in `README.md`.

I would just want to make sure it’s clear that this is an *optional* way of developing the project and not required :+1:

panoramix360:
Glad that you liked it!

I agree with you about this being optional.

I'll create the PR and change the README to include how to configure this :tada:

wmnnd:
This was implemented in #99 and is now part of the 0.9.0 release :partying_face:

PR #99, opened 2021-12-21
Adding DevContainer support

panoramix360:
As promised on issue #95, we added support for DevContainer from Microsoft on Visual Code.

We added some steps on README to help others running the project and marked as optional.


wmnnd:
Very cool, thanks for building this! :clap: 

I just gave it a try and I have a few thoughts:
- I needed to change the `docker-compose` file version to `"2"` to make it work on my machine. As far as I can see, it uses no features from 3.x
- It looks like the npm setup doesn’t quite work. When started the system for the first time, I needed to run `npm i` manually first, before it would work.
- The containers are all Debian-based and thus much larger than their Alpine counterparts. Changing this would probably require a fair bit of work, and I don’t know if it would be worth it. So I’d be happy to merge with the current setup :smiley:

panoramix360:
Nice points @wmnnd, thank you for letting me know.

I'll look into it to discover why `npm install` was not runned and the other points about the docker-compose version and Alpine.

I'll let you know my conclusions :)

panoramix360:
Hi @wmnnd, sorry for the delay in this change.

We are getting back from the holidays and will return our work here :D

wmnnd:
No worries, I’m very happy you’re still motivated to help :partying_face:

panoramix360:
@wmnnd we updated with what you suggested!

We included `mix deps.get` command on the postCreateCommand too, so after running devcontainer for the first time, the developer will only need to run `mix phx.server` :D

Please, let me known what you think

wmnnd:
Looking great, thanks for this contribution, it’ll make it super easy for new people who want to contribute :heart:

Issue #103, opened 2022-01-14
Add campaign data


wmnnd:
Campaigns, just like contacts, should allow for custom data in the shape of a `data` jsonb field.
This will allow the use of templates with Liquid to generate emails based on that data.

dpraul:
Very exciting to see this feature in the roadmap! Definitely going to follow this feature and product as you build it out. As you encouraged people to do, I just wanted to give my 2c on how this might be used in-practice.

We're currently using a lot of programmatic sugar around the MailChimp campaign data feature (via [their "editable areas" template language](https://templates.mailchimp.com/getting-started/template-language/)) to send data-driven newsletters on a regular basis. I do not necessarily think MailChimp is the best tool to do this, as we're definitely flexing it quite a bit, but it does the job.

MailChimp only allows the campaign data to be used via their template language, which is very weak (and afaik doesn't even exist in their newer template types), so we actually embed [handlebars](https://handlebarsjs.com/) _inside_ of the editable regions to make the templates more powerful. We stole this idea [from this Contentful blog post from 2018](https://www.contentful.com/blog/2018/11/30/mailchimp-campaign-contentful-webhooks-loose-coupling/). We chose handlebars simply because it is lightweight, but I have found we have to write lots of helpers to make it powerful enough to do everything we want.

We have a regular cronjob setup that pulls data from our API, applies it to a Mailchimp campaign, and sends it. The workflow, currently, is:

1. Kick off the cronjob by fetching a "base" campaign from MailChimp. 
    - We start it with a pre-existing campaign just so that administrators can use the MailChimp UI to easily adjust facets of the campaign to be applied to future emails (i.e. leverage MailChimp for what it's good for - the UI)
    - The campaign must be configured with an HTML template, which (as far as I know) is the only MailChimp template type that supports the campaign data feature
2. Create a new campaign using all of the settings of the prior
3. Fetch the default content for the editable fields of the campaign's template.
    - These fields are where we embedded the handlebars markup
4. Use handlebars to compile each of the editable fields with data from our API
5. Attach these compiled fields as the campaign data to the new campaign we created earlier
6. Send the campaign

It seems that the implementation as you've described it in [your twitter thread](https://twitter.com/pentacent_hq/status/1482027695512993795) would very easily fit nearly this exact workflow (and you've already built-in a much more powerful and friendly templating language!) - which is incredibly cool to see. I don't have much more to input since I haven't used keila at all, but I hope this input is useful in how you shape this feature!

[...]

PR #104, opened 2022-01-14
Add campaign custom data

wmnnd:
Implements #103 

Here is campaign data in action:
Screenshot 2022-01-14 at 17-19-51 Open Source Newsletter Tool · Keila

This PR does the following:

- Add new `JsonField` type for `campaign.data` and `contact.data`
- Adds `data` data field to campaigns
- Moves campaign settings to modal
- Moves update/send controllers to LiveView
- Adds preview button to WYSIWYG editor
- Removes toggle-wysiwyg button from Markdown campaigns (it’s been a bit buggy, so moving it to the settings modal discourages users from just going back and forth)


wmnnd:
@panoramix360, @gbottari: I’d really appreciate it if you could give this PR a look and let me know what you think! :pray:

wmnnd:
@panoramix360 Excellent! I’ve added some more details of what the PR changes in the PR description.

wmnnd:
I think there are some bugs in the editor in this branch, but I’m already working on fixing those.

panoramix360:
HI @wmnnd, I looked into your PR, sorry for the delay, the PR is very long hehe

Pretty nice feature! I think that will be good to have this on Keila! Congrats!

Some suggestions:
- Maybe it will be cool to add some reference sheet to make the user know about all the dynamic properties that he can add into the campaign email. Ex.: Contact first name, last name, etc
- It is not clear to me as user if I can use the custom fields from Campaign Data to fill the template.
- The refactoring to create a JsonField is pretty nice, it can be used in the future for other purposes.

I tested and found some minor bugs:
- I can scroll behind the preview when it is expanded.
- When I saved a campaign and edit again, I tried to open the preview and it was not showing.
- The preview is not reflecting the changes made on the WYSIWYG.
- I don't know if it was intended but when I inserted a incorrect JSON, the validation error was not really friendly.
Screenshot from 2022-02-07 20-41-57
- RememberUnsaved did not worked when editing the WYSIWYG field, I don't know if it's possible either hehe
- When I opened the Settings Dialog, I can type anything on Campaign Data field and close Settings clicking on the overlay background, if I open again the settings, the field will remain filled, maybe we can use the RememberUnsaved to solve this.
- When I opened the schedule Dialog and tried to send without filling any text fields, it was possible and did not show any validation error.

I noticed something strange too, I'll try to write the steps so you can reproduce:
- I created a campaign without having a Sender created.
- After creating the campaign and editing, I tried scheduling and it was informed that I needed a sender to send the campaign, so I think this behaviour is correct.
- Then I created a sender and opened the campaign
- When I clicked on the Settings, the sender was already selected for the campaign
- Then I clicked "Cancel"
- Tried to schedule again and received same error above
- I was only able to schedule when I opened settings and clicked on "Ok"

Maybe don't letting a campaign being created if the user has not configured any senders will be a better choice, I don't know.

Sorry for the extense reply haha

Tomorow, me and @gbottari will begin looking into the rate limiter issue.

But if you want I can help you tackle this issues above!

Let me know what you think :)

gbottari:
Hi guys! I think @panoramix360 made some great points. I also think this PR is great because it will enable the personalization of newsletters.

I noticed that `JsonField` does not allow JSON arrays. But it's not clear to me why that is. I think it is a valid use case. Imagine one wishes to reference a contact recommended products, like `['apples', 'oranges', 'bananas']`.

I also think it is worth limiting the maximum size of `contact.data` and `campaign.data` to prevent abuse (maybe adding a constraint).

wmnnd:
Hey @panoramix360 and @gbottari, thanks for this very useful feedback!

- Very good point about the discoverability of the contact and campaign data. Do you have any suggestions for improving that? Maybe a first step would be an additional page in the docs and linking them in the app.
- Good catch with the dialog scrolling and empty sender!
- I think allowing JSON arrays would make things a bit more complicated, because then it’s no longer guaranteed that `campaign.data` is always an object. If users want to use an array, they could easily just do something like `{"my_array": ["foo", "bar"]}`
- My latest commits have now already fixed the broken preview. I’ve also added improved Markdown editing while I was at it :smile: 
ezgif com-gif-maker


So here are the things that still need fixing as a takeaway from your comments:

- [ ] Make Liquid options more discoverable
- [ ] Disable scrolling background when preview is open
- [ ] Correctly reflect empty sender selection in campaign settings
- [ ] Add limit on contact and campaign data

panoramix360:
I'm glad that you liked our feedback, if you need any more help with this PR, just let us know

We'll now focus our efforts in the send rate issue

wmnnd:
I’m working on a new modal for adding Liquid tags, this should make things a bit easier! What do you think?

Screenshot 2022-02-10 at 22-26-19 Open Source Newsletter Tool · Keila

panoramix360:
I think this is a great opportunity to show the user what the feature can do!

wmnnd:
The Liquid dialog is now ready! Let me know what you think, I think this PR is ready to merge :partying_face:

panoramix360:
I tested and now I think it's pretty nice, well done

It'll be a great addition to the project :)

wmnnd:
@panoramix360 Thanks for testing it and all the feedback you’ve given, that’s been really helpful!

PR #118, opened 2022-02-28
added fix for style glitch in campaign settings Fixes #117

### LINKED ISSUE #117: Fix style glitch in campaign settings
If the sender name is particularly long, this breaks the layout of the campaign settings modal.

ExplorerKut:
Fix for style glitch in campaign settings 
Changed the files in :
lib/keila_web/templates/campaign/_settings_dialog.html.heex
and added sm:w-full class to the select element at line 45.
Before:

image

After:

image


wmnnd:
Awesome, thanks for giving it a shot! :heart: 

I think the `sm:` prefix is not needed for the class - because with it, the issue is only fixed on screens wider than 640px.

ExplorerKut:
also while going through different display sizes in devtools, I am encountering something like this, Is this something to be worried of?
image

wmnnd:
> also while going through different display sizes in devtools, I am encountering something like this, Is this something to be worried of?

You mean that the options are wider than the modal once you click on the select element? I wouldn’t worry about it because the Browser should take care that the options are displayed correctly.

ExplorerKut:
updated the PR, removed the sm prefix.

wmnnd:
Excellent, thank you!

PR #120, opened 2022-04-08
Expose default port (docker)

JDPDO:
If you are using an reverse proxy like [nginx-proxy-automation](https://github.com/evertramos/nginx-proxy-automation) containers have to be properly documented.

Exposing the default port adds the benefit that automated proxys and possible other tools are working out of the box.


wmnnd:
Thanks for your suggestion! 

Keila has two environment variables that can be configured for Keila: `PORT` and `URL_PORT`.
`PORT` is used as the port the application is listening on (usually `4000`) while `URL_PORT` is used as the port for generating public-facing URLs (usually 443 for HTTPS). Setting both variables to the same value may make sense in certain environments but won’t work in many other cases.

JDPDO:
You're right. Setting `URL_PORT` is too much.
Apart form that according to the [Dockerfile reference](https://docs.docker.com/engine/reference/builder/#expose) `EXPOSE` documents which ports the application serves. It not changes the behaviour of current compositions nor container runs. Furthermore it allows usage of `-P, --publish-all` flag.

wmnnd:
Your PR has now been merged, thank you! :heart:

Issue #119, opened 2022-04-07
Using html elements causing session breakdown 


JDPDO:
## What happens?

Using the specific character causing first some kind of session (dead-/live-)lock, then storage of corrupted entity, followed by internal server error (500). (_From future_: HTML enclosures are affected too.)

## How to reproduce

1. Create a new template, and name it as you want.
2. Type the sign ` (U+0060, grave accent) in signature field.
3. Things getting weird :confused:, let's get worse :smiling_imp:: Accept site reload and save the template.
4. Try opening the template causing internal server error.

wmnnd:
Thank you for reporting this issue. I haven’t been able to confirm it yet but I will take a closer look later.

Are you interested in helping create a fix for it?

JDPDO:
Yes, but I started elixir programming only several weeks ago.

JDPDO:
I've done some investigation. My foundings:

The error is caused by [earmark's parser](https://hexdocs.pm/earmark_parser/EarmarkParser.html) trying to parse an unfinished [_Gruber Markdown Text_](https://daringfireball.net/projects/markdown/syntax#backslash). The grave accent (\`) encloses for code formatting. For example complet copy <code>\`this works\`</code> and paste. No errors will occur.

What I do not understand so far is: Why are all other enclosures like \{ and \_ not affected?

JDPDO:
The error goes further than assumed: **All HTML element enclosures trigger an invalid template** the moment only one element added.

Try adding `<div>, <p>, ...` triggers the bug.

JDPDO:
What is if we are adding a function closing all open enclosures before passing text to earmarks parser?

JDPDO:
Investigating deeper:  Not using the rich text editor in campaigns edition triggers the bug too.

wmnnd:
The bug that caused session breakdowns with invalid markup in the signature field has been fixed in #175.

PR #125, opened 2022-04-11
Change way mails are build

JDPDO:
Changing the way mails are build.

- [x] Implement recipient behaviour
- [ ] Subdivide [`put_body`-method](https://github.com/pentacent/keila/blob/2c5fa08b943ccc5dd296f235682ce0ab0ab8f675/lib/keila/mailings/builder.ex#L137)
- [ ] Templates
  - [ ] Simplify signature insert
  - [ ] to structure e-mails in sections
  - [ ] to insert dynamic data
- [ ] Using AST the whole time to minimize parsing overheads. (Use Swoosh.Email.private map intended for frameworks.)
- [x] Prepare builder for #119: Flag for build function or second public function for live building.


wmnnd:
I’m closing this for now. If you still want to implement some of those features, please reach out and we can have a chat how to go about that :-)

Issue #107, opened 2022-01-18
Add rate-limiting to Senders


wmnnd:
Oban’s FLOSS version doesn’t allow to configure rate-limiting for queues, but it’s possible to use the [snoozing feature](https://hexdocs.pm/oban/Oban.Worker.html#module-snoozing-jobs) to implement rate-limiting with the help of another module like [ex_rated](https://github.com/grempe/ex_rated).

Rate-limits should be configurable per sender (or per shared sender) and be configured as `n` messages per unit of time (e.g. `second`, `minute`, `hour`, `day`).

It’s worth considering whether it should be possible to configure more than one rate limit (e.g. `100 messages per minute` *and* `1000 messages per hour`)

lucasoreis:
Hi @wmnnd! Me and @gbottari are thinking on work on this issue, is it okay for you?

Thank you!

panoramix360:
Me and @gbottari are thinking on work on this issue, is it okay for you?

Thank you!

panoramix360:
If you can share with us it'll be nice :D I think that will help

wmnnd:
Do you already have an idea how you want to implement the configuration of the rate-limits? Because a single sender might be subject to multiple limits (e.g. max 10 messages/second and max 100 messages/hour - or something like that).

gbottari:
I looked into ExRated. It seems like you can simply create a bucket for each sender using different message rates. However, controlling Oban's rate seems more involved because we need to do it in a way that won't trigger retries and timeouts. Do you have any thoughts on that?

wmnnd:
Yes, I think using ex_rated is a good idea.
We should probably add something like this to the Oban worker (just nicer, obviously :laughing:):

```elixir
with {:ok, _} <- ExRated.check_rate("#{sender.id}_second", 1000, sender.rate_limits["second"]),
     {:ok, _} <- ExRated.check_rate("#{sender.id}_minute", 60_000, sender.rate_limits["minute"])
       # …
      do
         # send email
else
  _ -> {:snooze, seconds}
end
```

I suppose we could ideally also calculate for how many seconds the job needs to be snoozed.

panoramix360:
@wmnnd this week I'll try to contribute to this issue!

Sorry for my absence in updating here :(

I'll try to tackle this one.

panoramix360:
Hi @wmnn :) 

Me and @gbottari added a simple implementation using ExRated.

[...]

PR #137, opened 2022-07-05
Adds a Rate Limit to Senders

panoramix360:
Implements the rate limit feature mentioned on issue #107

- Creates new fields for the Sender (`rate_limit_per_hour`, `rate_limit_per_minute`, `rate_limit_per_second`)
- Limits the calls in Oban using the ExRated library and snoozes the call if needed
- Updates UI to show new fields when creating a new sender


wmnnd:
Thanks for implementing this important feature, @panoramix360! :pray: 

I just have one suggestion for improvement: I think the new fields for the rate limits should be added to `SenderConfig` instead of `Sender`. Like this, they can be configured globally for shared senders which makes more sense I believe.

Let me know if you have time to look into this in the coming days - otherwise I can also take care of it :smiley:

panoramix360:
> Thanks for implementing this important feature, @panoramix360! pray
> 
> I just have one suggestion for improvement: I think the new fields for the rate limits should be added to `SenderConfig` instead of `Sender`. Like this, they can be configured globally for shared senders which makes more sense I believe.
> 
> Let me know if you have time to look into this in the coming days - otherwise I can also take care of it smiley

@wmnnd, I was looking into it but now I have some doubts.

Regarding the UI, do you think this needs to be configured per shared sender? Or the fields will be default for all senders?

Currently, they are like this:
Screenshot from 2022-07-22 12-45-17

So if we change per shared sender, these fields will appear inside each shared sender, is that it?

wmnnd:
The limit should be configurable for regular senders that don't use a shared sender - and for shared senders (where it will then be used for all senders that use this shared sender).

So it probably makes sense to add the UI fields after this line: https://github.com/pentacent/keila/blob/main/lib/keila_web/templates/sender/_config.html.heex#L52
Then they will be displayed among the other settings for each sender adapter.

panoramix360:
@wmnnd done!

Can you review and check the PR? :)

wmnnd:
Awesome, thanks for the update! I’ll review it this weekend.

wmnnd:
@panoramix360, @gbottari: Sorry for the delay, I’m on it!

panoramix360:
Hi @wmnnd! Did you have the chance to review it?

I saw that you added some modifications to keys and configs. Looks good.

wmnnd:
@panoramix360, @gbottari: I did some refactoring and testing and I think now everything is working as it should - both with regular senders and shared senders.

The major changes are:
- Rate limits are checked twice:
  - First, there's a preliminary check from large scale to small scale
     to allow short-circuiting when a bucket is already empty.
  - Secondly, the rate check goes from small scalle to large scale
     to hit buckets in the correct order.
- check_sender_rate_limit/1 has been moved to the Mailer module.
  It feels better to have it there as a companion function for
  deliver_with_sender/2 and shouldn't be in Sender because it has
  side-effects.

Let me know what you think! I think it’s ready to be merged now :partying_face:

panoramix360:
Hi @wmnnd,

Good additions to the PR, thank you!

I have just one small consideration:
- On line 27 in the file `ses_webhook_controller_test.exs`, I think that `@tag` is missing on the test.

Really nice that you created a TODO too, could be nice to make the algorithm better in the future!

That's it :D I really liked your suggestions and please let me know if you need anything more on this PR.

I'll look through the other issues to solve another :) If you have a suggestion, let me know!

wmnnd:
Thanks for noticing the missing tag! I think we’re ready to merge now :partying_face:

Issue #138, opened 2022-07-17
Upgrade to Oban 2.12


wmnnd:
This upgrade requires a migration, see https://hexdocs.pm/oban/v2-11.html

The new `Oban.Plugins.Reindexer` plugin was introduced as the result of a bug report from running Keila in production and should be enabled: https://github.com/sorentwo/oban/issues/612

wmnnd:
This is implemented in #139.

Issue #38, opened 2021-03-11
Image & File Uploads


wmnnd:
Right now, Keila has no way of including files or images with emails.

Before we start working on this, I think we should clarify the following questions:

1. Do we want to support inlining images and email attachments?
2. Which types of storage should be supported?
3. Should Keila support bring-your-own-storage?
4. Should Keila support processing (i.e. cropping/resizing) images?
5. Can we rely on available libraries to do some of the heavy-lifting?

And finally: Even if the answer to all of the above questions is Yes: Which of these are absolutely necessary for version 1.0?

wmnnd:
So my thinking is this:

1. No support for attachments and inlined images for 1.0.
2. Local file storage and maybe S3.
3. I don’t think it’s necessary to support bring-your-own-storage like we do for email senders. So one global storage configuration per instance should be enough.
4. I don’t see that as part of the MVP; it’s probably something we could add later once the templating system has a concept of "images".
5. I created a library called [Belt](https://hexdocs.pm/belt/Belt.html) to take care of sending files to different types of storage and to store the necessary information in the database. I created it almost four years ago, so I would certainly do some things differently now. But I do have it running in a production environment and it’s working quite well. On the other hand, if we were to support only local storage, we might not need something like this at all. I’m also open to consider alternative solutions or building something for Keila specifically.

I’d be very happy to hear some more thoughts :smile:

wmnnd:
P.S.: Maybe image resizing could also be handled client-side. It might be an option because with our default Markdown template, images wider than 600px make no sense.

cb3inco:
1. I think both attach and inline would be ideal
2. local and s3-compatible would be ideal - something configurable by the admin and its a either or option.
3.
4. Possibly basic editing or at least a warning if it's over a recommended size and the user can resize locally if needed.
5.

wmnnd:
This is how the first version of this feature will be implemented:

- Local storage only
- Support for resizing via imagemagick
- No cid or base64 support for the first version

wmnnd:
Implemented in #139.

PR #139, opened 2022-07-17
Add Image Uploads

wmnnd:
This pull request implements #38 in the new`Keila.Files` context.

Currently, only local file storage is supported, but serving those files from a different domain is possible (and encouraged).
There are currently also no image-specific features.

Browsing uploaded images is possible with a new LiveComponent file manager:

Screenshot from 2022-09-01 12-19-24

Implements #38 and #138.


wmnnd:
Added another small feature: Double-clicking on an image or clicking on the image button when an image is selected now allows modifying an existing image node.

wmnnd:
@panoramix360 & @gbottari: I think this PR is ready to go. Would you like to give it another look?

panoramix360:
@wmnnd nice!

I'll look into that :D

wmnnd:
@panoramix360  Did you have a chance to give the PR a look yet?

panoramix360:
Hi @wmnnd, sorry for the delayed answer!

I got a little caught up in my work, but I already got into working on an initial UI on code.

Didn't pushed it yet.

panoramix360:
Hi @wmnnd!

Sorry for the delayed answer, I just joined another company recently and was busy on onboarding and stuff haha

Looking through it now!

panoramix360:
@wmnnd 

Strange, it seems that the WYSIWYG editor is not opening here:
image

lib/keila/files/files.ex
```diff
@@ -0,0 +1,175 @@
...
+
+  @spec store_file(Project.id(), Path.t(), term) :: term | {:error, term()}
+  def store_file(project_id, source_path, metadata \\ []) do
+    filename = get_filename(source_path, metadata)
+    raw_type = Keyword.get(metadata, :type)
+
+    with {:ok, size} <- get_and_validate_size(source_path),
+         {:ok, type} <- get_and_validate_type(source_path, filename, raw_type) do
+      uuid = Ecto.UUID.generate()
+      sha256 = Keila.Hasher.hash_file(source_path, :sha256)
```
panoramix360, line 20: Maybe the type hash can be configured in `config.exs`, but I don't think it's really important

lib/keila_web/router.ex
```diff
@@ -156,6 +156,12 @@ defmodule KeilaWeb.Router do
     get "/c/:encoded_url/:recipient_id/:link_id/:hmac", TrackingController, :track_click
   end
 
+  scope "/uploads", KeilaWeb do
+    pipe_through :browser
+
+    get "/:filename", LocalFileController, :serve
```
panoramix360, line 162: Maybe on the future we could make this generic like `FileController` and send the file to Local storage or Remote storage if needed.

panoramix360:
Your code is really clean on this one :) everything is split into small modules and it seems extensible.

Since I don't use Elixir professionally yet, I really learned a lot of things by reading your code, so thank you for that! 😄 

Is there a possibility to the user/DevOps change the upload for a server/CDN folder?

It could be good to write something about it on the documentation, on how to create a new adapter to store files remotely. What do you think?

Wrapping up, I think it's pretty good update on Keila 💯

wmnnd:
Thank you so much for taking a look and for your kind words @panoramix360!

I’ve set up the code in such a way that it’s theoretically relatively easy to add another file storage adapter for, say, S3, but I think that’s something that can be added if there’s demand for it and someone who wants to take it on :smile: 
I’ll take a look at your other notes tomorrow.

lib/keila_web/router.ex
wmnnd: That was originally my idea, but I think generally it’s probably preferable to use the generated URLs for including files in emails rather than serving them through Keila.

lib/keila/files/files.ex
wmnnd: The SHA256 hash is also stored in the database (though it’s not currently used in any way … but). In the future this might be useful for features like deduplication though - or for searching for known files in the storage.

PR #145, opened 2022-09-06
Makes system from email configurable

wmnnd:
Implements #48.


aqua-institut:
Great PR! That would solve the problem we have encountered while setting up Kyla on-prem.

PR #164, opened 2022-11-25
Adding User-Agent filtering

panoramix360:
Adds User-Agent filtering to open links accordingly to this [blog post](https://www.gmass.co/blog/false-opens-in-gmail/).

I used the follow User-agent for the filter:
`Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/42.0.2311.135 Safari/537.36 Edge/12.246 Mozilla/5.0`

@wmnnd, I'm creating the PR before completing the tests so you can take a look beforehand :smile: 

Please let me know what you think.

### DIFF
--- a/lib/keila/tracking/tracking.ex
+++ b/lib/keila/tracking/tracking.ex
@@ -156,10 +156,17 @@ defmodule Keila.Tracking do
     end
   end
 
-  def track(:open, %{encoded_url: encoded_url, recipient_id: recipient_id, hmac: hmac}) do
+  def track(:open, %{
+        encoded_url: encoded_url,
+        recipient_id: recipient_id,
+        hmac: hmac,
+        user_agent: user_agent
+      }) do
     case verify_hmac(hmac, encoded_url, recipient_id) do
       :ok ->
-        set_recipient_opened_at(recipient_id)
+        unless is_user_agent_bot(user_agent) do
+          set_recipient_opened_at(recipient_id)
+        end
 
         {:ok, URI.decode_www_form(encoded_url)}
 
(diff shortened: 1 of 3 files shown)


wmnnd:
Awesome, this is looking great! I’m just wondering whether we might want to store the User-Agent alongside the event, in order to potentially find further bot opens. But on the other hand, that is a slight privacy compromise. What do you think?

panoramix360:
Hmm, understood.

Do you mean saving the event on Database the user-agent used for opening the link?

Do you think this is a breach of privacy? I was searching about it and found that WordPress does this too if I'm right hehe

I'm okay with saving it for now. Are you thinking of using it to try to detect more bots in the future, using recurrence or something like that?

panoramix360:
@wmnnd I'm finishing the tests and I notice that the documentation for some methods seems a little odd.

The @doc for the open event mentions the `link_id` as optional, but in the method there is no `link_id` or it should be?

```elixir
def track(:open, %{
        encoded_url: encoded_url,
        recipient_id: recipient_id,
        hmac: hmac,
        user_agent: user_agent
      }) do
    case verify_hmac(hmac, encoded_url, recipient_id) do
      :ok ->
        unless is_user_agent_bot(user_agent) do
          set_recipient_opened_at(recipient_id)
        end

        {:ok, URI.decode_www_form(encoded_url)}

      :error ->
        :error
    end
  end
```

wmnnd:
@panoramix360 You’re right, `link_id` is not used by the open event.

wmnnd:
Thank you for your contribution, @panoramix360! The PR looks good and it's a great addition to Keila to filter out those requests :+1:

PR #179, opened 2023-01-11
Add a Block Editor based on Editor.js

wmnnd:


wmnnd:
It’s finally here! :partying_face: 

@everyone: I would like to merge this PR next week, and would be very grateful for feedback before that - maybe there are still some bugs I haven’t yet managed to find :lady_beetle: 

keila-block-editor

wmnnd:
@panoramix360 & @gbottari As always, your feedback would be especially appreciated :blush:

wmnnd:
> I'll look into it this weekend, sorry for not doing anything for these months!

That would be fantastic! :hugs:  Looking forward to hearing what you think of the new editor!

assets/js/campaign-editors/block/blocks/button/index.js
panoramix360, line 1: I don't know what you think, but we could add [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) to the HTMLs being created like this for example:

```
drawView() {
  this.wrapper.innerHTML = "";
  this.wrapper.className = this.data.centered ? "flex flex-col items-center" : "";

  this.wrapper.innerHTML = `
    <div class="button-contenteditable" style="margin:0!important;" contenteditable="true">
      ${this.data.label}
    </div>
    <input type="url"
           placeholder="https://www.example.org"
           class="text-xs mt-1 w-full bg-transparent border-1 border-gray-500 border-dashed"
           value="${this.data.url}">
  `;

  this.button = this.wrapper.querySelector(".button-contenteditable");
  this.linkEditor = this.wrapper.querySelector("input[type='url']");
}
```

panoramix360:
I'm taking a look piece by piece, it's a very long PR haha

wmnnd:
@panoramix360 What do you think, is the PR ready to be released? :sweat_smile:

assets/js/campaign-editors/block/index.js
```diff
@@ -0,0 +1,91 @@
...
+          inlineToolbar: true
+        },
+        separator: Separator
+      },
+      onChange() {
+        window.setUnsavedReminder(true)
+        editor.save().then((outputData) => {
+          source.value = JSON.stringify(outputData)
+          source.dispatchEvent(new Event("input", { bubbles: true }))
+        }).catch((error) => {
```
panoramix360, line 71: Maybe we should deal with this kind an error? if it's not possible to save the block

assets/js/campaign-editors/block/index.js
```diff
@@ -0,0 +1,91 @@
...
+          class: Header,
+          config: {
+            levels: [1, 2, 3]
+          }
+        },
+        layout: {
+          class: Layout,
+          config: {
+            tools: {
+              header: {
```
panoramix360, line 28: This header object could be above, because it's repeated on line 19 :)

assets/js/campaign-editors/markdown/index.js
```diff
@@ -0,0 +1,87 @@
...
+          view.dom.parentNode.parentNode.querySelector(`.wysiwyg--menu [data-action=img]`).dispatchEvent(
+            new Event("mousedown")
+          )
+        }
+      }
+    })
+  }
+
+  get content() {
+    return markdownSerializer.serialize(this.view.state.doc)
```
panoramix360, line 66: If this throws an error what can happen?

Maybe encapsulating the `serialize` inside a try/catch could be a good idea

assets/js/campaign-editors/markdown/menu.js
```diff
@@ -0,0 +1,272 @@
...
+      e.stopPropagation()
+      e.preventDefault()
+      editorView.focus()
+      const activatedItem = items.find(({ dom }) => dom.contains(e.target))
+      if (activatedItem) activatedItem.exec(editorView)
+    }, true)
+  }
+
+  update() {
+    this.items.forEach(item => item.update(this.editorView.state))
```
panoramix360, line 39: The editorView.state will always be available?

assets/js/campaign-editors/markdown/menu.js
```diff
@@ -0,0 +1,272 @@
...
+  }
+
+  destroy() {
+    this.dom.remove()
```
panoramix360, line 43: It could be good to remove the "click" and "mousedown" eventListener when the MenuView is destroyed.

assets/package.json
```diff
@@ -1,11 +1,18 @@
 {
   "repository": {},
   "description": " ",
-  "license": "MIT",
+  "license": "agpl-3.0",
```
panoramix360, line 4: Interesting that you changed the license hehe
I'm curious

lib/keila/mailings/builder.ex
```diff
@@ -65,7 +67,11 @@ defmodule Keila.Mailings.Builder do
 
   defp put_template_assigns(assigns, _), do: assigns
 
-  defp process_assigns(value) when is_number(value) or is_binary(value) or is_nil(value) do
+  defp process_assigns(value) when is_number(value) or is_binary(value) do
```
panoramix360, line 70: These two next functions are not doing the same thing?

Why not group the `when` guards?

lib/keila/mailings/builder.ex
panoramix360, line 1: This module is doing a lot, maybe we should create another one to deal with template stuff.

lib/keila/templates/html_text_extractor.ex
```diff
@@ -0,0 +1,140 @@
+defmodule Keila.Templates.HtmlTextExtractor do
+  @doc """
+  Processes an HTML fragment tree and returns it as a string.
+  """
+  @spec extract_text(Keila.Templates.Html.t(), String.t()) :: String.t()
+  def extract_text(html, selector \\ "body") do
+    Floki.find(html, selector)
+    |> do_extract_text(0)
+    |> concat_extracted_text()
+    |> then(fn string -> string <> "\n" end)
```
panoramix360, line 10: In this line, if you change to this:
`<> "\n"`

It will have the same result, no?

lib/keila/templates/html_text_extractor.ex
```diff
@@ -0,0 +1,140 @@
...
+  defp do_extract_text({"hr", _attrs, _content}, _list_level) do
+    "\n---\n\n"
+  end
+
+  defp do_extract_text({"a", attrs, content}, list_level) do
+    href = html_attribute(attrs, "href")
+    title = html_attribute(attrs, "title")
+    content = do_extract_text(content, list_level)
+
+    case title do
```
panoramix360, line 71: This case could be clearer with this:

```
if title do
    ["[", content, "]", ~s{(#{href} "#{title}")}]
  else
    ["[", content, "]", "(#{href})"]
  end
```

lib/keila/templates/html_text_extractor.ex
```diff
@@ -0,0 +1,140 @@
...
+  end
+
+  defp do_extract_text(text, _) when is_binary(text) do
+    whitespace_left = if String.match?(text, ~r{^\s}), do: :whitespace
+    whitespace_right = if String.match?(text, ~r{\s$}), do: :whitespace
+
+    [whitespace_left, String.trim(text), whitespace_right]
+  end
+
+  defp concat_extracted_text(list) do
```
panoramix360, line 98: This function could be broken in:
- Separate function to deal with whitespace
- Separate function to deal with new lines

panoramix360:
I made a lot of comments :)

They are just suggestions feel free to let me know what you think!

I never worked with Liquid template system so it's cool to take a look haha

I'm sorry for taking so long to review, I'll try to be better in my next responses.

It's nice to be back looking through Elixir code 😃

lib/keila/templates/html_text_extractor.ex
wmnnd: I’m not sure we could actually do this because in the context of HTML, line breaks are usually treated like white spaces. There are actually kinda complex rules around when newlines are whitespace and when they are simply ignored; I tried to implement a good general-purpose solution here that works for most cases.

lib/keila/templates/html_text_extractor.ex
wmnnd: I think it’s mostly a matter of preference here^^ Erlang folks would probably prefer the `case`, people from other backgrounds the `if`.

lib/keila/templates/html_text_extractor.ex
wmnnd: It would work without the then, but then the `<> "\n"` wouldn’t be part of the pipe anymore and it would mess up the formatting :sweat_smile:

lib/keila/mailings/builder.ex
wmnnd: I agree! I think this is a task for a future refactoring for sure.

lib/keila/templates/html_text_extractor.ex
panoramix360: You are right 😅

wmnnd:
@panoramix360 Thank you for adding your comments, I think you’re making some very good points!

At this point I feel like I want to ship the new feature as soon as possible and then do some refactoring at the same time as users are already trying out the new experience. Have you been able to play around with the new features from a user side a bit and have you found any blockers (from a user perspective) there?

lib/keila/templates/html_text_extractor.ex
panoramix360: Great then! :D

panoramix360:
@wmnnd no problem :D

I didn't have the time to play around as a user, I just looked through the code.

But you can merge the push and I'll look after it, if any bugs or things appear I'll create an issue then, what do you think?

assets/js/campaign-editors/block/index.js
wmnnd: Good point, maybe just show an alert that indicates something has gone wrong instead of failing silently.

lib/keila/mailings/builder.ex
wmnnd: Good point! I think it evolved organically like that, but they can be/should be grouped.

assets/js/campaign-editors/markdown/menu.js
wmnnd: The event listeners are attached to `this.dom`, so I think they should be removed automatically when the node is remomved, no?

wmnnd:
> But you can merge the push and I'll look after it, if any bugs or things appear I'll create an issue then, what do you think?

Sure, that would be great! Though I’m not gonna merge it before next week, because I’m currently at ElixirConf EU :smiley:

panoramix360:
> > But you can merge the push and I'll look after it, if any bugs or things appear I'll create an issue then, what do you think?
> 
> Sure, that would be great! Though I’m not gonna merge it before next week, because I’m currently at ElixirConf EU 😃

Oh nice!! :D

Are you showcasing the project there? How is it going?

assets/js/campaign-editors/markdown/menu.js
panoramix360: Yes, they are, but there are some cases in which we might run into memory leaks.

Especially, if the element is referenced in another place or an anonymous function or closure. So, just for safety, it's best to remove the event listeners before removing the element from the DOM.

lib/keila/mailings/builder.ex
wmnnd: I’m tracking this issue here: #206

lib/keila/mailings/builder.ex
wmnnd: They have been grouped now.

assets/package.json
wmnnd: The license didn’t actually change - Keila has always been AGPLv3 from the start, but the package.json originally created by the Phoenix generator didn’t reflect that.

assets/js/campaign-editors/markdown/menu.js
wmnnd: I tried moving the functions out but they are scoped to the constructor scope, so it wasn’t possible to do it just like that. This PR also doesn’t actually change this code, it’s just moved to a different file :-)

assets/js/campaign-editors/markdown/menu.js
wmnnd: Yep, it comes from Prosemirror.

assets/js/campaign-editors/block/index.js
wmnnd: I added a console warning here.

Issue #182, opened 2023-01-18
Default `MAILER_SMTP_PORT` not 587 as docs suggest


aej:
The docs say that the default `MAILER_SMTP_PORT` is 587. However keila does not handle a default. If `MAILER_SMTP_PORT` is unset then it will set the value to `nil`

PR #184, opened 2023-01-18
Set mailer smtp port default to 587

aej:
Resolves https://github.com/pentacent/keila/issues/182


wmnnd:
Thanks for noticing and fixing this!

PR #195, opened 2023-03-01
add Sendinblue

vinnpinn:
This PR adds the Sendinblue API to the supported senders. 
The UI might need some tweaking since the available senders are now displayed on two rows, however the UI isn't broken:
image


wmnnd:
Hey @Vinnpinn, thank you for this PR! I will review it next week and I’d be happy to add Sendinblue as an additional sending API!

Vinnpinn:
Great, we've been using it with our own instance, the API for sending mails seems to work without any issue.
The webhooks for handling bounces and complaints are missing. I had a look at how it's implemented for SES, however I'm not familiar with elixir or erlang at all, so the syntax is confusing me a lot. 
Not sure if I'm able to implement the webhooks, I might have another look when I have some free time, however not in the near future.

Issue #205, opened 2023-04-24
Feature request: add support for FriendlyCaptcha


beeb:
[FriendlyCaptcha](https://friendlycaptcha.com/) is a nice and privacy-preserving alternative to hCaptcha that doesn't require tedious actions by the user.

I would to see this as an option for subscription forms in Keila. What's more, FriendlyCaptcha doesn't set cookies which makes it easier to embed into GDPR-compliant webpages.

beeb:
I could try! It's not so different from hcaptcha so I guess I could fight my way through it with copy pasting even though I don't know the language. I'll give it a go when I have some time.

beeb:
@wmnnd how would you see this being implemented from the user's perspective? Instead of a checkbox, it would be a `<select>` when creating a form, with two options? What about the database, how is this stored at the moment?
What would be required if the database schema changes? I'm not sure how migrations are handled etc.

That's a bit much for me to take on as a first elixir project ever I think.

wmnnd:
I think the best way to implement this is by making it entirely configurable with environment variables. I don’t think it’s necessary/useful for users to be able to select among multiple captcha providers from the interface.
Nothing about the captcha is stored in the database except for the boolean on the form that enables/disables them.

beeb:
@wmnnd trying to run the current state of `main` inside the dev container like described in the contribution guide, but compilation fails with:

```
== Compilation error in file lib/keila/templates/default_template.ex ==
[...]

PR #209, opened 2023-05-02
Add support for FriendlyCaptcha

beeb:
[FriendlyCaptcha](https://friendlycaptcha.com/) is a nice and privacy-preserving alternative to hCaptcha that doesn't require tedious actions by the user.

This PR adds support for FriendlyCaptcha as an alternative to hCaptcha. The choice is done by providing the relevant format for the site key environment variable.

If the site key contains a hyphen character, indicating a UUID format, hCaptcha will be chosen. Otherwise, FriendlyCaptcha will be used.

All references to hCaptcha have been renamed to Captcha where relevant. Notably, the following environment variables should be used for configuration:

- `CAPTCHA_SITE_KEY`
- `CAPTCHA_SECRET_KEY`
- `CAPTCHA_URL`

For backwards compatiblity, the current variables (`HCAPTCHA_SITE_KEY`, etc) are checked in case the above variants are not provided.

Documentation should probably be updated to reflect this. The default value for the API url when using FriendlyCaptcha is `https://api.friendlycaptcha.com/api/v1/siteverify`.

This PR closes https://github.com/pentacent/keila/issues/205


config/config.exs
```diff
@@ -65,11 +65,12 @@ config :keila, Keila.Billing,
...
 
-# Staging configuration for hCaptcha
-config :keila, KeilaWeb.Hcaptcha,
+# Staging configuration for hCaptcha or FriendlyCaptcha
+config :keila, KeilaWeb.Captcha,
   secret_key: "0x0000000000000000000000000000000000000000",
   site_key: "10000000-ffff-ffff-ffff-000000000001",
-  url: "https://hcaptcha.com/siteverify"
+  url: "https://hcaptcha.com/siteverify",
+  type: "hcaptcha"
```
wmnnd, line 73: The type should be an atom (`:hcaptcha`)

config/runtime.exs
```diff
@@ -88,27 +88,43 @@ if config_env() == :prod do
...
-  hcaptcha_url = System.get_env("HCAPTCHA_URL")
+  # Captcha
+  captcha_site_key = System.get_env("CAPTCHA_SITE_KEY") || System.get_env("HCAPTCHA_SITE_KEY")
 
-  if hcaptcha_site_key not in [nil, ""] and hcaptcha_secret_key not in [nil, ""] do
-    config =
-      [secret_key: hcaptcha_secret_key, site_key: hcaptcha_site_key]
-      |> put_if_not_empty.(:url, hcaptcha_url)
+  captcha_secret_key =
+    System.get_env("CAPTCHA_SECRET_KEY") || System.get_env("HCAPTCHA_SECRET_KEY")
```
wmnnd, line 95: Nice approach for backwards compatibility!

config/runtime.exs
```diff
@@ -88,27 +88,43 @@ if config_env() == :prod do
...
+
+  captcha_url = System.get_env("CAPTCHA_URL") || System.get_env("HCAPTCHA_URL")
+
+  if captcha_site_key not in [nil, ""] and captcha_secret_key not in [nil, ""] do
+    captcha_type = if captcha_site_key =~ "-", do: "hcaptcha", else: "friendlycaptcha"
```
wmnnd, line 100: I don’t think this method is suitable for determining the captcha type. I suggest you use an additional `CAPTCHA_PROVIDER` env variable for that.

config/config.exs
wmnnd: Also, let’s maybe call it `provider` instead of `type`.

config/runtime.exs
```diff
@@ -88,27 +88,43 @@ if config_env() == :prod do
...
+    Captcha not configured.
     Keila will fall back to using hCaptcha’s staging configuration.
 
-    To configure hCaptcha, use the following environment variables:
+    To configure a captcha, use the following environment variables:
 
-    - HCAPTCHA_SITE_KEY
-    - HCAPTCHA_SECRET_KEY
-    - HCAPTCHA_URL (defaults to https://hcaptcha.com/siteverify)
+    - CAPTCHA_SITE_KEY (or HCAPTCHA_SITE_KEY for compatibility purposes)
```
wmnnd, line 125: I think we can drop the reference to the `HCAPTCHA` env variables here because we don’t want to encourage their use anymore.

config/runtime.exs
```diff
@@ -88,27 +88,43 @@ if config_env() == :prod do
...
+    System.get_env("CAPTCHA_SECRET_KEY") || System.get_env("HCAPTCHA_SECRET_KEY")
+
+  captcha_url = System.get_env("CAPTCHA_URL") || System.get_env("HCAPTCHA_URL")
+
+  if captcha_site_key not in [nil, ""] and captcha_secret_key not in [nil, ""] do
+    captcha_type = if captcha_site_key =~ "-", do: "hcaptcha", else: "friendlycaptcha"
+    Logger.info("Using the #{captcha_type} captcha provider")
 
-    config :keila, KeilaWeb.Hcaptcha, config
+    default_captcha_url =
```
wmnnd, line 103: In Elixir, people generally prefer case statements for such situations:

```elixir
default_captcha_url =
  case captcha_provider do
    :hcaptcha -> "https://hcaptcha.com/siteverify"
    :friendly_captcha -> "https://api.friendlycaptcha.com/api/v1/siteverify"
  end

File: config/runtime.exs (hunk 3 above)
Comment (wmnnd) [reply config/runtime.exs:100]: So I think you should do something like this:

```elixir
captcha_provider =
  System.get_env("CAPTCHA_PROVIDER", "hcaptcha")
  |> String.downcase()
  |> case do
    "friendly_captcha" -> :friendly_captcha
    _other -> :hcaptcha
  end
end
```

lib/keila_web/controllers/form_controller.ex
```diff
@@ -32,15 +32,23 @@ defmodule KeilaWeb.FormController do
     end
   end
 
+  defp captcha_response_param() do
+    if captcha_config()[:type] == "hcaptcha" do
```
wmnnd, line 36: This should be a `case` statement.

lib/keila_web/helpers/captcha/captcha.ex
```diff
@@ -0,0 +1,74 @@
...
+      true
+    else
+      _other -> false
+    end
+  end
+
+  defp request_body(response) do
+    config = config()
+
+    if config[:type] == "hcaptcha" do
```
wmnnd, line 48: `case` statements, please, here and in the other functions of this module :wink:

wmnnd (commented):
Hey, thank you for this PR! Great job, especially given that this is your first time using Elixir. 

I’ve left some comments, most of them regarding code conventions in Elixir, like how `case` is more typically used than `if` in many instances.

config/runtime.exs
beeb: Good idea!

beeb:
Thanks for the very insightful comments! I went through and fixed the code accordingly.

beeb:
@wmnnd let me know if anything else is needed for this to be merged. Thanks

satoshinotdead:
How can I try this patch if I'm using docker version of Keila?

Sorry because my 'noobish' question but that will be my first time doing that way (privacy always matters for us) :)

wmnnd:
Hey @beeb, thanks again for contributing this PR and for implementing the changes I proposed in my review. I have changed some things around a little bit to keep all captcha-related logic in the captcha module. Other than that, great job!

If you want, it would be great if you could also update the website docs at https://github.com/pentacent/keila-web to make sure the new way of configuring captchas is visible to everyone :smiley:

beeb:
Done!  https://github.com/pentacent/keila-web/pull/8
😄

PR #216, opened 2023-06-07
Ordering Segments Alphabetically

panoramix360:
This PR changes the query that gets all segments and order results alphabetically using Ecto.

image


wmnnd:
Awesome, thank you! And nice touch adding a test case :+1:

Issue #141, opened 2022-07-24
Feature request : implement double opt-in


MagicFab:
I'd like to suggest a default feature found in other similar software : an option to request a new email subscriber to confirm their email is valid by clicking on a link sent in the first confirmation email after subscription. This is commonly known as double opt-in.
This improves a lot the quality of email subscribers by minimizing typing mistakes while also reducing SPAM, specially when not using the CAPTCHA method.

This could be an optional, but checked-by-default checkbox under the option "Require captcha" option at form creation.

https://www.campaignmonitor.com/resources/glossary/double-opt-in/

wmnnd:
Thanks for creating this issue. You’re right, double-opt is an absolute must and I’ll add it to the roadmap.

I wonder whether this is best done via configuring the form (i.e. you set the form to double-opt-in and then also configure the message you’d like to send) - or if this should be part of the automation feature #132.

Happy to hear everyone’s thoughts!

victorbjorklund:
Agreed. This is a must have. Also would be nice to have access to it when adding with the api (if you use your own form instead of the form in keila).

wmnnd:
This is how I think the double-opt in feature should work:

- Add a new tab to the form editor called "Double-opt-in" with the following elements:
  - Checkbox: Enable double-opt-in
  - Select: Sender
  - Select: Email template
  - Text field: Email subject
  - Text area: Markdown editor for editing the opt-in content (no WYSIWYG for the first iteration)
    -  This field validates the inclusion of the `{{ opt_in_link }}` variable.
  - Text field: Redirect URL
- Contacts that sign up from a form with double-opt-in should be stored in a separate table, ideally with their email address encrypted/obfuscated.
- Once the opt-in link is clicked, a regular contact is created

wmnnd:
This feature was implemented in #219 and is available in Keila 0.13.0  :partying_face:

PR #219, opened 2023-07-03
Implement Double-Opt-In

wmnnd:
This PR implements #141 and adds a double-opt-in feature that can be configured for every signup form.


tcurdt:
Very nice - but for compliance the IP and time of the consent must be saved.

wmnnd:
Hey, thanks for your feedback. The time when the double-opt in is confirmed is already saved. You can see it when you click on a contact in your contact list.

I haven’t seen anything in German/European regulations that suggests, storing the IP address is in any way necessary. I know it’s considered *legal* to store IP addresses during the DOI process, but not *mandatory*.

In fact, you might argue that the way Keila handles DOI is much more in line with the spirit of the law: Keila does not create an entry in your contact list at all before DOI has been completed. The proof that a user has consented to receiving your newsletter ultimately comes from the fact that they have used the personalized opt-in link that was sent to their email.

wmnnd:
@tcurdt You might find this useful reading (in German): https://socialmediarecht.wordpress.com/2012/12/05/das-urteil-des-olg-munchen-az-29-u-168212-zum-double-opt-in-verandert-nichts/

tcurdt:
That would be great but it sure sounds mandatory here:

https://www.e-recht24.de/artikel/ecommerce/6534-newsletter-rechtssicher-erstellen-und-versenden.html

I guess there are two different interpretations that would need to be tested in courts.
What a mess.

wmnnd:
Seems like the lawyers disagree here, but as far as I’m aware there has been no decision by a court or even a memo from a regulatory body that suggests collecting IP addresses would be required or even useful. Because of this, I’d err on the side of caution and try to collect as little data as possible :blush:

tcurdt:
...but does the time of the confirmation get stored? I didn't see that either yet.

wmnnd:
Yep, the UTC timestamp is stored in a dedicated field `double_opt_in_at` for the contact.

Issue #110, opened 2022-02-10
Allow download of contact list as CSV


wmnnd:
The Contacts page should allow downloading contacts as CSV.
The CSV file should be formatted to be compatible with CSV import files.

The same download feature should also be available from the segment editor.

In order to allow also large contact lists to be downloaded, `Keila.Contacts.stream_project_contacts/2` should be used
with `Plug.Conn.send_chunked/2` and `Plug.Conn.chunk/2`.

MagicFab:
This feature is particularly important to avoid problematic cloud vendor lock-in : you can migrate your lists easily into Keila but can't migrate out once in as is right now - unless you're self-hosting and/or have direct access and knowledge to extract from the database. Data portability is essential to reassure users they have 100% control and ownership of their data - and their email recipients'.

satoshinotdead:
I wonder if there is a good practice as manual workaround in order to export the database.

_All tables from postgres and that's it?_

wmnnd:
This feature is now available in version 0.12.6 :partying_face:

MagicFab:
Awesome ! Thank you :1st_place_medal:

PR #238, opened 2023-09-21
Add contacts CSV export capability

katafrakt:
This adds possibility to export contacts from a project or a segment to CSV, compatible with CSV import. It addresses #110.

I wasn't really sure where to put the buttons, so please advise for a better place.

## Screenshots

screenshot-localhost_4000-2023 09 22-00_30_27

screenshot-localhost_4000-2023 09 22-01_15_06

Screenshot_20230922_011119

## Notes

The code in these two controllers look very similar. I wasn't sure if it makes sense to abstract it out and especially - where. This feels like highly controller-level code, so pushing it to a context does not feel right. There could be a shared codebase space for just controllers, but I didn't want to just create it without consultation.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -8,7 +8,8 @@
 import Config
 
 config :keila,
-  ecto_repos: [Keila.Repo]
+  ecto_repos: [Keila.Repo],
+  csv_export_chunk_size: 100
 
 # Configures the endpoint
 config :keila, KeilaWeb.Endpoint,
--- a/config/config.exs
+++ b/config/config.exs
@@ -7,9 +7,9 @@
 # General application configuration
 import Config
 
-config :keila,
-  ecto_repos: [Keila.Repo],
-  csv_export_chunk_size: 100
+config :keila, ecto_repos: [Keila.Repo]
+
+config :keila, KeilaWeb.ContactsCsvExport, chunk_size: 100
 
 # Configures the endpoint
 config :keila, KeilaWeb.Endpoint,
--- a/config/test.exs
+++ b/config/test.exs
@@ -14,7 +14,9 @@ config :keila, Keila.Repo,
   timeout: 60_000,
   pool_size: 16
 
-config :keila, skip_migrations: true, csv_export_chunk_size: 3
+config :keila, skip_migrations: true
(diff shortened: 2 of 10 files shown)


wmnnd:
This is awesome, thank you for creating this PR, @katafrakt! :partying_face: 

Here are some small changes to the code I’d suggest:

1. In most interfaces in Keila, all action buttons are in the top row, so I suggest you put the export button on the contacts and segment pages up there as well, probably in the left-most position.
2. Like you’ve already mentioned, the code for both controllers is pretty much the same. I think it makes sense to create a helper module for centralizing this code. My suggestion would be adding `KeilaWeb.ContactsCsvExport` in `lib/keila_web/helpers` with a function `stream_csv_response/3` that takes as its first argument `Plug.Conn` and the arguments for `Contacts.stream_project_contacts/2` as its second and third argument.
3. I also recommend you move the chunk configuration away from being directly under the `:keila` key to something like `config :keila, KeilaWeb.Helpers.CsvExport, chunk_size: 100`

katafrakt:
@wmnnd  I made the changes you suggested.

screenshot-localhost_4000-2023 09 24-20_49_58

screenshot-localhost_4000-2023 09 24-20_50_27

I'm not 100% sure about the location of the download button in the segment page. It is very detached from the contacts list. And, until the segment will be saved, it will not reflected the changes done by the user. Now I'm thinking that maybe that button should be even placed here? 

screenshot-localhost_4000-2023 09 24-20_52_45


Not sure though, it's your project and you have the vision for it, I'm just sharing some thoughts.

wmnnd:
@katafrakt Can you give me write permissions to your branch so I can adjust the button positions? Then I’ll merge your PR :smiley:

katafrakt:
@wmnnd you should already have it. At least that's how I always understood this option.

screenshot-github com-2023 10 01-20_44_57

wmnnd:
Ah, you’re right. Looks like this just didn’t work with Github’s github.dev editor.

wmnnd:
Thank you for your contribution, @katafrakt! I have just released a new version that includes mainly your new feature :partying_face:

Issue #284, opened 2024-04-23
Allow creating users from administer users page


doobry-systemli:
Trying out the official keila Docker image, I didn't find a way to add new users when logged in as root user. The web interface only lists the sole user, without any options to add/edit/delete users:

![grafik](https://github.com/pentacent/keila/assets/25278864/e9a8cd4d-2684-4f23-9e88-72ff57f5bd5c)

Do I miss something? Or is user management somehow disabled in the official Docker image?

wmnnd:
Hey, thank you for your question! You can’t currently create new users on this page. You can register new users at `/auth/register`

doobry-systemli:
Oh, I see. So there is no option to disable user registration but at the same time manage users as the admin user?

doobry-systemli:
We're thinking about running a  keila instance for activist projects, but open registration is no option for us. So we'd envision a setup where our admin user can add new users manually on request and these users can maintain their campaigns and contact sets.

wmnnd:
You can quite easily accomplish this via the console at the moment before this feature will eventually be added. 

Open the console like this on the container:

`/opt/app/bin/keila remote`

Then execute the following command:

```elixir
Keila.Auth.create_user(%{email: "user@example.com", password: "secure-password-123"})
```

wmnnd:
I’ve updated the title of the issue and added it to the roadmap. We should add a feature to create new users from the admin page.

pmareke:
Hi @wmnnd I would like to give it a try.

The change should be enable a button to create users, right? Do you want to list/edit them as well?

Do you have any sketch about the UI?

Thanks!

wmnnd:
This has now been implemented by @pmareke! 🥳

PR #332, opened 2024-08-26
Create user as admin, solves #284

pmareke:
This pull request solves #284  and adds the possibility to create new user from the UI  if the user is admin.

- [X] Add new button to create using in the users admin panel.
- [X] Add a new form to create users.
- [X] Validate the fields in the form using LiveView.
- [X] Add translations.
- [X] Add tests for the new UserAdminController#create method.

SCREENSHOTS:

- New button:
image

- Empty form:
image

- Live errors:
image

- All fields ok:
image

- New user added:
image

I try to follow the same form as in another pages.

### DIFF
--- a/lib/keila_web/live/create_user_live.ex
+++ b/lib/keila_web/live/create_user_live.ex
@@ -1,0 +1,36 @@
+defmodule KeilaWeb.CreateUserLive do
+  use KeilaWeb, :live_view
+  alias Keila.Auth.User
+
+  @impl true
+  def mount(_params, _session, socket) do
+    socket =
+      socket
+      |> assign(:changeset, Ecto.Changeset.change(%User{}))
+
+    {:ok, socket}
+  end
+
+  @impl true
+  def render(assigns) do
+    Phoenix.View.render(KeilaWeb.UserAdminView, "new.html", assigns)
+  end
+
+  @impl true
+  def handle_event("validate", params, socket) do
+    changeset =
+      %User{}
+      |> User.creation_changeset(params["user"])
+      |> then(fn changeset ->
+        if changeset.valid?,
+          do: changeset,
+          else: Ecto.Changeset.apply_action(changeset, :update) |> elem(1)
+      end)
+
+    socket =
+      socket
+      |> assign(:changeset, changeset)
+
+    {:noreply, socket}
+  end
+end
--- a/lib/keila_web/live/create_user_live.ex
+++ b/lib/keila_web/live/create_user_live.ex
@@ -19,13 +19,9 @@ defmodule KeilaWeb.CreateUserLive do
   @impl true
   def handle_event("validate", params, socket) do
     changeset =
-      %User{}
-      |> User.creation_changeset(params["user"])
-      |> then(fn changeset ->
-        if changeset.valid?,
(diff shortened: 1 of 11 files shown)


wmnnd:
Oh nice, thank you for building this feature, I’m sure it’ll make a lot of people happy (including me)!

I’ll do a review review of your changes on the weekend or next week 😊

lib/keila_web/live/create_user_live.ex
```diff
@@ -0,0 +1,36 @@
...
+
+  @impl true
+  def handle_event("validate", params, socket) do
+    changeset =
+      %User{}
+      |> User.creation_changeset(params["user"])
+      |> then(fn changeset ->
+        if changeset.valid?,
+          do: changeset,
+          else: Ecto.Changeset.apply_action(changeset, :update) |> elem(1)
```
glhrmv, line 27: Can we change the action to `:validate` just so that it's clearer this is to trigger error helper rendering?

wmnnd: @glhrmv `:validate` is not an action that’s used by Ecto. https://hexdocs.pm/ecto/Ecto.Changeset.html#t:action/0

wmnnd: It’s actually not necessary to use `apply_action/2` here at all. The docs suggest simply setting the `action` field directly: https://hexdocs.pm/ecto/Ecto.Changeset.html#module-changeset-actions

So this whole assignment could be simplified like this:

```elixir
changeset = 
  params["user"]
  |> User.creation_changeset()
  |> Map.replace!(:action, :insert)
```

glhrmv: Hey @wmnnd thank for the response - I'm a greenhorn Phoenix developer so I really appreciate the links to the Ecto docs.

However, are you sure that code snippet is equivalent to what the MR is trying to do? It seems like the `:action` is being set only when the changeset is *invalid* - I thought the reason for this was to provide helpful error validation messages alongside the inputs in the UI, since Phoenix form error helpers only show errors if `:action` has a value, as per this section in the docs: https://hexdocs.pm/phoenix_live_view/1.0.0-rc.6/Phoenix.Component.html#form/1-a-note-on-errors

wmnnd: The original code applies the action using `apply_action` which returns `{:ok, data}` if valid or `{:error, changeset}` if invalid. Note that the original code then returns the original changeset if `{:ok, data}` was returned and the updated changeset (with the action set) if `{:error, changeset}` was returned. So in both cases, we end up with a changeset. If the changeset was valid in the original code, we end up with a changeset without `action`, but for a valid changeset it makes no difference anyways.

pmareke: Thank you both, I'll change as @wmnnd mentioned!

wmnnd (approved):

wmnnd:
Thanks for adding this long-awaited feature, @pmareke, great job!

pmareke:
> Thanks for adding this long-awaited feature, @pmareke, great job!

Thanks to you for let me help!

PR #339, opened 2024-10-11
Pass url_fn when an admin creates a new user

hellehata:
When a new user is created in the admin section the link in the activation email does not include a URL.


wmnnd:
Thank you for catching the fact this was missing! I’m merging it now 😊

Issue #356, opened 2024-12-04
Switch MAILER_SMTP_FROM_EMAIL and MAILER_SMTP_USER defaults


wmnnd:
Currently, only `MAILER_SMTP_USER` needs to be set. If it's set, `MAILER_SMTP_FROM_EMAIL` defaults to the value of `MAILER_SMTP_USER`. This logic should be reversed to avoid issues where the SMTP user is not an email address or considered private.

kevinam99:
Hey @wmnnd. I decided to take this up and opened the PR for it here https://github.com/pentacent/keila/pull/358

kevinam99:
Just a nudge to close this issue

PR #358, opened 2024-12-18
Switch MAILER_SMTP_FROM_EMAIL and MAILER_SMTP_USER defaults and fix CLA signing script

kevinam99:
Aims to fix issue https://github.com/pentacent/keila/issues/356 with the same title

Does this require any tests to be updated?


kevinam99:
Also fixed the issue with the CLA signing script

wmnnd:
Oh, fabulous, thanks for taking this on! I’ll include this in the next release and update the website docs accordingly.

Also thanks for signing the CLA and reporting the issue with the `base64` tool. It looks like there is an incompatibility between the macOS version (which you were probably using) and the GNU versions. While the macOS version has the `-b` parameter, GNU versions have `-w` But I’ll fix that after merging your PR.

kevinam99:
Yeah, probably a macOS thing so I went ahead and removed the flag altogether. That too seems to achieve the same result

Issue #289, opened 2024-05-03
reduce log level to error via env variable


tcurdt:
When running the docker image I would like to avoid logging the simple http requests.
I was hoping I could just set a ENV variable and reduce logging from INFO to ERROR - but I couldn't find any docs about this yet.

I bet there is a way?

wmnnd:
Thanks you for suggesting this! I’ll add this to our roadmap.

kevinam99:
Is anyone doing this? I can take this up if it is available

kevinam99:
Opened the PR here https://github.com/pentacent/keila/pull/360 
Let me know if it is okay :)

PR #360, opened 2024-12-19
Set log level via env variable

kevinam99:
Aims to fix https://github.com/pentacent/keila/issues/289


wmnnd:
Thanks for taking this on! Could you also allow `debug` to the allowed log levels and change the name of the env variable to just `LOG_LEVEL`?

kevinam99:
Done!

wmnnd:
Excellent, thanks!

PR #367, opened 2025-01-05
Delete account feature

### LINKED ISSUE #353: Account deletion feature
As far as I can tell, a feature to delete the Keila account is currently missing.
While this is not very important for self-hosted instances, this should still exist, especially for the (or possibly a) hosted version.

danilax86:
Added `DELETE /account` method and updated account page.

Should we allow account removal for root users?

Closes https://github.com/pentacent/keila/issues/353

image
image


azamaziz989:
Wow! Thats good. I thing account deletion should also have email confirmation.

wmnnd:
Hey @danilax86, thanks for this contribution! The reason I haven’t reviewed it properly yet is that I'm not sure yet how I want account deletion to be handled on Keila Cloud. This solution is probably going to be fine for self-hosted instances, but I still need to give it some more thought.
I’ll get back to this PR after January 20 😊

danilax86:
Hi @wmnnd! Any updates yet?

Issue #368, opened 2025-01-05
Creating an API key with empty name causes weird message


marc-bouvier:
![image](https://github.com/user-attachments/assets/ebd6010a-fed2-4f04-ab59-afd7b6634426)

![image](https://github.com/user-attachments/assets/13256fd9-57d5-4568-aae3-b80a253cce8e)

wmnnd:
Thanks for reporting this! This was some leftover inspection code from debugging 😅

PR #369, opened 2025-01-06
Remove token's data inspection when creating API token without a name

danilax86:
Inspection is definitely great for debugging, but, I guess, there is no need to show the whole structure for users this way.

Closes: https://github.com/pentacent/keila/issues/368


danilax86:
image

You can create API token without a name and it should be totally fine if we use an ID of token as its name.

wmnnd:
Oh, looks like this was left over from debugging this feature. Thanks for fixing!

PR #381, opened 2025-02-07
remove hardcoded hello@keila.io address for amazon ses verification mail

steam0r:
this solves pentacent/keila/issues/370 by removing a hardcoded email address

tried to make it close to what is being done in `emails.ex`, hope thats fine


wmnnd:
Thanks for taking a look at this! However, I don’t think this is the right way to go about this.

If you want to be able to customize the verification From email, that should happen in the settings for the Shared Sender. It doesn't really make sense to tie it to the system mailer.

steam0r:
> Thanks for taking a look at this! However, I don’t think this is the right way to go about this.
> 
> If you want to be able to customize the verification From email, that should happen in the settings for the Shared Sender. It doesn't really make sense to tie it to the system mailer.

i just tried to make it somehow work, adding this to the settings of the shared sender seems like a good idea,
i was trying to mimic what the default sender does, sorry if this does not make much sense. but with the hardcoding
in it currently is not possible to use sas at all...or i am not getting it.

wmnnd:
The system mailer (which is used to send things like the registration confirmation email or password reset emails) is configured with environment variables because there can only be one system mailer.

Shared Senders, on the other hand, are configured through the interface. A single instance of Keila could have multiple shared senders. Because of this, everything related to them, should be configured in the interface and not with an environment variable - just like all other properties of the shared sender (like server, password, etc).

steam0r:
would it then be possible to simply use the from address here, or does this need an extra field?

Bildschirmfoto 2025-02-10 um 14 00 18

wmnnd:
Oh, so you're trying to just set up a regular SES Sender, not using a Shared Sender?

steam0r:
so, i think setting up a "non shared" sender with ses works. when i set up a shared sed sender and use that in a project, the frontend tells me:

Bildschirmfoto 2025-02-13 um 19 12 40

and then results in this:

Bildschirmfoto 2025-02-13 um 19 13 14

and this logentry on the server:

Bildschirmfoto 2025-02-13 um 19 13 36

which, i guess, is the server sending with a hardcoded from-email that is not verified at SES...and can't be.
so i searched for the hardcoding and looked at what the other sender so and hence my PR.

am i doing something wrong?

Issue #377, opened 2025-01-22
Align devcontainer configuration to version changes


marc-bouvier:
To reproduce

Either
- Open the repository in Github Codespaces

<img width="534" alt="Image" src="https://github.com/user-attachments/assets/6d89097d-e259-47a2-94de-44555155a411" />

- Checkout and open the repository locally, then choose "reopen in container"

<img width="623" alt="Image" src="https://github.com/user-attachments/assets/ae9fb967-cb3b-4a37-b989-3462eeaad1ff" />

### Elixir version mismatch

```text
[{
	"resource": "/workspace/mix.exs",
	"owner": "_generated_diagnostic_collection_name_#3",
	"severity": 8,
	"message": "an exception was raised:\n    ** (Mix.ElixirVersionError) You're trying to run :keila on Elixir v1.14.4 but it has declared in its mix.exs file it supports only Elixir ~> 1.15\n        (mix 1.14.4) lib/mix/tasks/deps.loadpaths.ex:60: Mix.Tasks.Deps.Loadpaths.check_elixir_version/1\n        (mix 1.14.4) lib/mix/tasks/deps.loadpaths.ex:39: Mix.Tasks.Deps.Loadpaths.run/1\n        (mix 1.14.4) lib/mix/task.ex:421: anonymous fn/3 in Mix.Task.run_task/4\n        (mix 1.14.4) lib/mix/tasks/loadpaths.ex:41: Mix.Tasks.Loadpaths.run/1\n        (mix 1.14.4) lib/mix/task.ex:421: anonymous fn/3 in Mix.Task.run_task/4\n        (mix 1.14.4) lib/mix/tasks/compile.ex:127: Mix.Tasks.Compile.run/1\n        (mix 1.14.4) lib/mix/task.ex:421: anonymous fn/3 in Mix.Task.run_task/4\n        (language_server 0.26.2) lib/language_server/build.ex:441: ElixirLS.LanguageServer.Build.run_mix_compile/1\n        (language_server 0.26.2) lib/language_server/build.ex:80: anonymous fn/3 in ElixirLS.LanguageServer.Build.build/3\n        (stdlib 4.3.1) timer.erl:235: :timer.tc/1\n        (language_server 0.26.2) lib/language_server/build.ex:20: anonymous fn/3 in ElixirLS.LanguageServer.Build.build/3\n        (kernel 8.5.4) global.erl:477: :global.trans/4",
	"source": "Elixir",
	"startLineNumber": 1,
[...]

PR #392, opened 2025-03-08
Update dev container dependencies

henriqueberlesi:
When building the project locally, I've faced two issues with the current setup:
- `fast_html` requires cmake as a dependency
- `keila` requires Elixir 1.15 or above

This PR adds `cmake` to the devcontainer image, and bumps Elixir version in `docker-compose`


wmnnd:
Thank you for updating the devcontainer settings! I’m not really using it myself so I never notice when something breaks but it's great if it's useful for other people 😊 

Issue #377 mentions that the devcontainer should also include `drpint`. Could you add that as well? Then I’m gonna merge your PR.

henriqueberlesi:
Sure @wmnnd. After a few attempts, I've installed `dprint` in the Dockerfile and then moved it to `/usr/local/bin` to make it available to the `vscode` user.

Running `dprint` works when inside a terminal, and the output from the issue changed to:
```
[INFO] Initialized dprint 0.49.0
  Folder: /workspace/
  Command: dprint
[INFO] Extension active!
```

I believe we're good to go :muscle:
Let me know if there's any other change needed.

wmnnd:
Excellent, thank you!

Issue #390, opened 2025-03-06
Mark emails as unreachable if delivery fails because they are invalid


wmnnd:
Currently, if a contact email is invalid (e.g. `foo@bar-.#`), delivering to them will quietly fail. This should be surfaced by marking them as unreachable.

wmnnd:
This was implemented in #393.

PR #393, opened 2025-03-08
Mark emails as unreachable

henriqueberlesi:
My attempt on #390.

Updating the `handle_result` from `Keila.Mailings.Worker` to deal with the already existent `:invalid_email` return error. The action will update the `failed_at` date for the recipient, and update the recipient's contact to the `unreachable` atom.

Also, refactored a few queries into their own private functions. Maybe in the future these functions could be moved to their own service/module?

Also, @wmnnd: I have not found any tests covering this file. Should we add tests with this PR?

Thanks!


wmnnd:
Hey @henriqueberlesi, thanks you for taking on this issue! I’ll take a closer look in the coming days 😊

lib/keila/mailings/worker.ex
```diff
@@ -106,19 +107,70 @@ defmodule Keila.Mailings.Worker do
   # Email was already sent
   defp handle_result({:error, :already_sent}, _), do: {:cancel, :already_sent}
 
+  # Email is invalid
+  defp handle_result({:error, :invalid_email}, recipient) do
+    Multi.new()
```
wmnnd, line 112: Could you refactor this to just

```
Repo.transaction(fn ->
  recipient
  |> set_recipient_failed_query()
  |> Repo.update_all([])

  # ...
end)

We don’t really use Ecto.Multi anywhere else in the codebase and so I think overall this is easier to read 😊

wmnnd:
Sorry it took so long, but overall this PR is looking good and is a valuable contribution, thank you! I just left one small comment – and once that’s been addressed, I’d be happy to merge 🚀

henriqueberlesi:
@wmnnd thanks for the feedback! I refactored `Ecto.Multi` out, as requested. Please let me know if there are any other changes needed.

wmnnd:
Excellent, thank you! Sorry for the nit-pick about Ecto.Multi 😅

PR #405, opened 2025-03-22
Added Message Stream setting to Postmark adapter

codecat:
This adds an option for [Postmark's Message Streams](https://postmarkapp.com/support/article/1207-how-to-create-and-send-through-message-streams) to the Postmark adapter. (Different types of message streams go through different Postmark infrastructures, so this seemed like a useful option to have.)

This is implemented using [the Swoosh Postmark provider option](https://github.com/swoosh/swoosh/blob/e878e49f78d960f9c60bc85abcfad38e56cba80a/lib/swoosh/adapters/postmark.ex#L57-L58) for message streams.

Tested it, and it seems to work.

(Disclaimer: This is my first time ever touching Erlang - hope I did it right!)


wmnnd:
Hey, thank you for creating this PR, it’s looking good to me!

Since I can’t test this Postmark setting myself, would you be able to confirm you tested this setting locally and that it’s working as expected?

codecat:
Yes, it works locally! 👍

image

When left empty it will use the default, which is the transactional stream.

wmnnd:
Awesome, thanks for confirming that it’s working! I’ve now merged your PR 🚀

Issue #312, opened 2024-07-05
Timestamps in the logs show no date


almereyda:
The logs contain hour, minute, second and increments for actions logged to the console.

Unfortunately they don't contain a date, which renders it harder to correlate events, in case no additional log postprocessing is happening.

As an administrator, I often need to look up log output, in order to correlate it with human activity.

It will be helpful to have a date as a hint, which allows to focus more easily.

wmnnd:
Thank you for suggesting this. I agree, this would be helpful and I’m adding it to the roadmap.

geniuskidkanyi:
I would love to contribute to this, are you still looking for help?

wmnnd:
Sure, please feel free to open a PR that updates the Logger format to include a full ISO 8601 timestamp!

wmnnd:
This was implemented in #432 .

PR #432, opened 2025-06-14
Include ISO 8601 timestamp

samuelc92:
Example:
`2025-06-14T21:35:28.126`

I think that Logger format doesn't provide support for full ISO 8601 out of the box. Therefore, they way I implemented doesn't contain the timezone indicator in the end of the timestamp

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -75,7 +75,7 @@ config :keila, KeilaWeb.Captcha,
 
 # Configures Elixir's Logger
 config :logger, :console,
-  format: "$time $metadata[$level] $message\n",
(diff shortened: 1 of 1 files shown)


wmnnd:
The docs for [Logger.Formatter](https://hexdocs.pm/logger/main/Logger.Formatter.html) mention there are features for timezone conversion but then it doesn't actually mention them again (as far as I can see) 😅 
But then again it's probably not that important to include the timezone since people should be aware which timezone their server is running on.

This implements #312.

samuelc92:
Agree

Issue #436, opened 2025-06-17
Allow SES for system mailer


andreialecu:
It appears that there needs to be a system mailer defined, which can only be SMTP: https://www.keila.io/docs/configuration

It would be nice if SES could be used here as well.

wmnnd:
Hey there, thank you for your suggestion!

You can use SES via SMTP so I don't think there's a need for this.

andreialecu:
You can, and I set it up that way. But it requires enabling SMTP for SES, which is an additional step and potential security issue.

PR #438, opened 2025-06-20
Improved user management

andreialecu:
This adds many things:

- the ability for the root user to edit users, and update their name, email, activate/deactivate them if they cannot confirm the email that gets sent out,
- adds optional email 2FA, a code is sent to the user on login that they must enter to log in
- adds optional 2FA with WebAuthn (Face ID/Touch ID)
- many tests are included

It has been mostly written with the help of Claude Sonnet 4, but under supervision. I have tested it manually as well and things seem to work just fine.


andreialecu:
Video walkthrough of changes:


https://github.com/user-attachments/assets/0beeba55-06ca-4831-86d1-9888d150cbc6

wmnnd:
Hey there, thank you for this PR! Like with the other PR, I'd ask you to break it into smaller bits.

I think the most straightforward part is improving the admin page and allowing the root user to change a user's email, name, and resend the activation email. Not entirely sure if it makes sense to allow manually changing the status - when would you want to do that?
Changing the password and sending the forgot password link is probably also a good addition.

I can't accept the 2FA changes right now because I'll have to read up on best practices for 2FA first and this will take some time. Just some thoughts: The main method I'd want to support is TOTP, so if we add 2FA, that must be supported. Then there is usually some sort of fallback code.

andreialecu:
I realize it's a bit of a big PR - TOTP can be easily added, but with WebAuthn, I feel that TOTP is deprecated nowadays.

I mainly use it to be able to create users without an actual email behind the scenes. You could for example create users like `joe@localhost` or whatever other invalid email.

It was particularly useful during development. But it seems that the root user should be able to override activation status, in case of support requests, for example. Perhaps they are not receiving the email, the SMTP server is down, or whatever other reason. It's only something the root user can do.

Also I do not have a valid SMTP server to use for the system mailer and would rather not expose one, so it was a workaround to be able to confirm users. (see #436)

wmnnd:
Hey @andreialecu, I want to say thank you again for your initiative - but like I said I can‘t accept PRs this size. I will look at 2FA best practices later this year 😊

PR #483, opened 2025-12-15
Fix quotes in data column of ODS import template

individual-it:
some quotes in the LibreOffice template to import contacts are wrong

`{“marketing_consent”: true, “age”: 34, “birthplace”: “Paris”, "tags":["rocket-scientist"]}`
should be
`{"marketing_consent": true, "age": 34, "birthplace": "Paris", "tags":["rocket-scientist"]}`

note the `“` vs `"` in some places

and same in the second example line.

with the existing quotes the import fails

BTW: thank you for that great piece of software!


wmnnd:
Thank you for fixing this, this is indeed a frequent source of problems!

Issue #465, opened 2025-09-30
Full height editor?


simonrepp:
Hi there!

When I write the text of a campaign I'm always somewhat hindered by the side-by-side input and preview only spanning half the height of my browser viewport - with the lower half just sitting there unused and empty - forcing me to scroll around much more and have less of an overview of what I'm doing. Likewise, although I should know better by now, I regularly and instinctively reach out to drag the textarea field to occupy the full height (which works) - but as soon as I enter a single character in the editor, the height collapses down to half-height again (which feels very frustrating at some point :)).

Is there any chance the editor can be tweaked to make use of the full viewport height? For me personally 95% of what I do with keila is be on that page to write the newsletter, so for me this would be a big improvement (and I don't really see any drawbacks from implementing it, maybe it's even trivial depending on your existing CSS).

And: Thanks so much for keila, it's super cool!

<img width="1709" height="1325" alt="Image" src="https://github.com/user-attachments/assets/ddee4507-8b1d-450c-a424-6912ca4d159e" />

wmnnd:
Thank you for suggesting this, I think this is an excellent idea and we should add it! It probably requires a little bit of JS but maybe I can look into it when I have some free time 😊

simonrepp:
Excellent, thank you, looking forward to it! :)

Safal2004:
Hi! I would like to work on this issue if it is still available.
Could you please assign it to me?

wmnnd:
Hey there, there's already a PR by @RohilPrajapati (#488) for which I've added some changes today (so that it automatically gets larger with the content). If any of you have feedback on this before I merge it, let me know!

PR #488, opened 2025-12-21
feat(campaign-editor): use full viewport height for editor #465

RohilPrajapati:
- Remove textarea rows attribute
- Add Tailwind height utilities for full-height layout


wmnnd:
Hey @RohilPrajapati, sorry it too me so long to look into this - I finally found some time today and added a little Alpine code to automatically resize the Markdown editor according to the content. Let me know what you think!

RohilPrajapati:
Hello, @wmnnd  look good then just max height 1080 px but if we could add synchronous scroll to both preview and editor that would be great UX.

wmnnd:
Thank you again for creating this PR and suggesting this feature! I have decided to implement it slightly differently as part of the big template refactor and the new implementation is part of Keila 0.30.0 which is about to be released!

PR #499, opened 2026-01-17
Add Hungarian translation

trueqap:
## Summary
- Add Hungarian (hu) locale support to Keila
- Add Hungarian to the language selector dropdown
- Complete Hungarian translation for all domains:
  - `admin.po` - Admin interface strings
  - `auth.po` - Authentication strings
  - `cloud.po` - Cloud feature strings
  - `default.po` - Main UI strings
  - `errors.po` - Error messages

## Test plan
- [x] Build and run Keila locally with Docker
- [x] Verify Hungarian appears in language selector (Account → Change Language)
- [x] Verify UI displays correctly in Hungarian after switching language

### DIFF
--- a/priv/gettext/hu/LC_MESSAGES/auth.po
+++ b/priv/gettext/hu/LC_MESSAGES/auth.po
@@ -28,22 +28,22 @@ msgstr "Aktiválás sikeres"
 #: lib/keila_web/templates/auth/reset_success.html.heex:4
 #, elixir-autogen, elixir-format
 msgid "Check your inbox!"
-msgstr "Ellenőrizd a beérkező leveleidet!"
+msgstr "Ellenőrizze a beérkező leveleit!"
 
 #: lib/keila_web/templates/auth/login.html.heex:34
 #, elixir-autogen, elixir-format
 msgid "Don’t have an account?"
-msgstr "Nincs fiókod?"
+msgstr "Nincs fiókja?"
 
 #: lib/keila_web/templates/auth/login.html.heex:40
 #, elixir-autogen, elixir-format
 msgid "Forgot your password?"
-msgstr "Elfelejtetted a jelszavad?"
+msgstr "Elfelejtette a jelszavát?"
 
 #: lib/keila_web/templates/auth/activate_success.html.heex:12
 #, elixir-autogen, elixir-format
 msgid "Get started!"
-msgstr "Kezdjük!"
+msgstr "Kezdés!"
 
 #: lib/keila/auth/emails.ex:46
 #, elixir-autogen, elixir-format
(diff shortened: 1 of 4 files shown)


wmnnd:
Hey @trueqap, thank you for contributing this translation! 🇭🇺 

Before I can merge this PR, I would kindly ask you to sign the CLA. You can read how that works here:
https://github.com/pentacent/keila/blob/main/.cla/README.md
Please use your real name for signing it; your data is stored in an encrypted file and stays private so no one else can read it 😊

trueqap:
Thanks\! I've signed the CLA. 🙂

trueqap:
Thanks for the quick merge\! I'll maintain the Hungarian translation going forward. 🙏

wmnnd:
Thanks! Since I don't speak a word of Hungarian, I asked an LLM for a second opinion and it pointed out the following:

- There seems to be a random Russian word in one of the strings: `"A megadott fájlt nem удалось feldolgozni."`
- It pointed out that there might be a slightly more elegant way for the plural forms of which there seem to be several instances.
  ```
  msgid "should have %{count} item(s)"
  msgstr[0] "%{count} elem(ek)kel kell rendelkeznie"
  msgstr[1] "%{count} elem(ek)et kell tartalmaznia"
  ```
  Instead, it suggests this:
  ```
  "%{count} elemmel kell rendelkeznie"` / `"%{count} elemet kell tartalmaznia"
  ```
- The LLM also found that you're using the informal address throughout (which is good, I'm also doing that for the German translation) but it said that there are a couple of strings that use the formal address.

So ... no idea how valid these points are, but maybe you could take a look and let me know what you think 😊

wmnnd:
> Thanks for the quick merge! I'll maintain the Hungarian translation going forward. 🙏

It's not merged yet 😉

trueqap:
Hi! I noticed an error in the meantime - I accidentally submitted an earlier version of the translation.

The fix has now been pushed. The main change is using formal addressing ("Ön" form) instead of informal ("te" form), which is more appropriate for a professional application.

Sorry for the confusion!

wmnnd:
Now that I think of it, Keila already uses different registers across translations. German, Spanish, and Bulgarian use the informal address, and French uses the formal address. Out of curiosity, I just checked the website of another email tool. They use the informal address for German, Spanish, and Italian but formal address for Portuguese and French. 

So if it makes more sense to use the formal address for you as a speaker of Hungarian, let's do it this way!

trueqap:
I've been maintaining the Hungarian WordPress translation for 20 years, and there we use the `_formal` suffix to distinguish between the two versions (e.g., `hu_HU` for informal and `hu_HU_formal` for formal).

Perhaps we could introduce the same convention in Keila? I'd be happy to create both versions.

The reason I went with the formal version now is that the project I was translating this for is a government contract, so formal addressing was more appropriate. Though personally, I actually prefer the informal style! 😊

wmnnd:
Maintaining two versions sounds like a lot of work 😅 
I will merge this version now though, thanks again for contributing it! Next week I'll probably cut a new release and the translation will be included.

PR #509, opened 2026-02-15
Fix/#501 - broken shared sender admin

### LINKED ISSUE #501: Missing _config.html template breaks shared sender admin
## Bug Description

When trying to create a new shared sender at `/admin/shared-senders/new`, the following error occurs:

```
** (Phoenix.Template.UndefinedError) Could not render "_config.html" for KeilaWeb.SenderView
```

## Root Cause

In commit `de0f3932` ("Refactor sender editing to use LiveView"), the file `_config.html.heex` was renamed to `_config_new.html.heex`:

```
.../{_config.html.heex => _config_new.html.heex}
```

However, the `shared_sender_admin/edit.html.heex` template was not updated and still references the old `_config.html`:

```elixir
# lib/keila_web/templates/shared_sender_admin/edit.html.heex:60
<%= render(KeilaWeb.SenderView, "_config.html", form: f, sender_adapters: @sender_adapters) %>
```

## Steps to Reproduce

1. Log in as admin
2. Navigate to `/admin/shared-senders/new`
3. Server error occurs

## Suggested Fix

Either:
1. Create a new `_config.html.heex` template for shared senders that uses `render_shared_sender_adapter_form/2`
2. Or update `shared_sender_admin/edit.html.heex` to use a different approach

I can submit a PR with the fix if helpful.

BrutalBirdie:
Fixes #501

- [X] Singed the CLA
- [X] Followed the [CONTRIBUTING.md](https://github.com/pentacent/keila/blob/main/CONTRIBUTING.md)
- [X] Tested the PR locally with the dev container setup

> [!NOTE] 
> Dev Container - currently, this is set:
> https://github.com/pentacent/keila/blob/fc1fb46bf53bcb804e2e354a37266d1179eee4c9/.devcontainer/docker-compose.yml#L25
> Which results in an error:
> ```
> Error: in 18+, these Docker images are configured to store database data in a
>        format which is compatible with "pg_ctlcluster" (specifically, using
>        major-version-specific directory names).  This better reflects how
>        PostgreSQL itself works, and how upgrades are to be performed.
> 
>        See also https://github.com/docker-library/postgres/pull/1259
> 
>        Counter to that, there appears to be PostgreSQL data in:
>          /var/lib/postgresql/data (unused mount/volume)
> 
>        This is usually the result of upgrading the Docker image without
>        upgrading the underlying database using "pg_upgrade" (which requires both
>        versions).
> 
>        The suggested container configuration for 18+ is to place a single mount
>        at /var/lib/postgresql which will then place PostgreSQL data in a
>        subdirectory, allowing usage of "pg_upgrade --link" without mount point
>        boundary issues.
> 
>        See https://github.com/docker-library/postgres/issues/37 for a (long)
>        discussion around this process, and suggestions for how to do so.
> ```
> Had to set it to `image: postgres:17` to get the dev container working.

### DIFF
--- a/lib/keila_web/templates/sender/_config.html.heex
+++ b/lib/keila_web/templates/sender/_config.html.heex
@@ -1,13 +1,18 @@
 <%= inputs_for @form, :config, fn fc -> %>
   <% tab = input_value(fc, :type) || List.first(@sender_adapters) %>
-  <% tab_data = "{ tab: '#{tab}', senderType: $el.querySelector('#shared_sender_config_type').value }" %>
+  <% tab_data =
+    "{ tab: '#{tab}', senderType: $el.querySelector('#shared_sender_config_type').value }" %>
 
   <div
     class="tabs"
     x-data={tab_data}
     x-init="$nextTick(() => { if (! $el.querySelector('button.tab-label.active')) {$el.querySelector('button.tab-label').click()} })"
   >
-    {select(fc, :type, @sender_adapters, id: "shared_sender_config_type", x_model: "senderType", class: "hidden")}
+    {select(fc, :type, @sender_adapters,
+      id: "shared_sender_config_type",
+      x_model: "senderType",
+      class: "hidden"
+    )}
 
     <%= for adapter <- @sender_adapters do %>
       <% class = "{ 'active': tab === '#{adapter}' }"


aj-foster:
Hello 👋🏼  I may be a bit confused, but why does this PR introduce a new template? It looks like the desired template already exists, it was simply renamed.

BrutalBirdie:
Hello @aj-foster 
Yes, nice that there is a new template, but the new template is not beeing used, but the old one.
Thus trying to create a shared sender throws an error.

Keila Version 0.19.0 deployed just to answer your comment:
```
Mar 22 10:08:46 2026-03-22T09:08:46.166 request_id=GJ8esY3tIGWZ13EAABex [error] ** (Phoenix.Template.UndefinedError) Could not render "_config.html" for KeilaWeb.SenderView, please define a matching clause for render/2 or define a template at "lib/keila_web/templates/sender/*". The following templates were compiled:
```

Like it is documented in issue #501, which this PR is trying to solve in a simply way.
Perhaps there already has been changes made in the main branch since this PR and the issue was commented.

> Same issue with [`0.17.1` ](https://github.com/pentacent/keila/releases/tag/0.17.1), [`0.18.0`](https://github.com/pentacent/keila/releases/tag/0.18.0) and latest commit [`fc1fb46bf53bcb804e2e354a37266d1179eee4c9`](https://github.com/pentacent/keila/tree/fc1fb46bf53bcb804e2e354a37266d1179eee4c9).

that might fix this.
But the current release is still bricked.

wmnnd:
Hey @BrutalBirdie and @aj-foster, thank you both for looking into this issue! I have now merged #515 because it makes more sense to use the updated template instead of duplicating the old template.

PR #520, opened 2026-04-10
Add option to disable tzdata auto-update check

sachabertschibfs:


wmnnd:
Thank you, that's looking good! Even if this is just a small change and might not technically require it, could you please also sign the CLA? You can find the instructions for that here: https://github.com/pentacent/keila/blob/main/.cla/README.md

wmnnd:
Thank you!

PR #521, opened 2026-04-10
feat: add TZDATA_AUTOUPDATE env var to disable tzdata autocheck

### LINKED ISSUE #519: Disabling tzdata autocheck through environment variable
Hello,

A second issue I faced when running Keila through RHOS (RedHatOpenShift) is that tzdata try to check if there's an update.
To do so, it need to write into an internal txt file. This is blocked in RHOS and create an error in the logs of the pod.
tzdata tries again, every few second, so it create a LOT of logs.

It would be nice to have an environment variable to disable that.

From what I found you need to add this to deactive it in runtime.exs :
`config :tzdata, :autoupdate, :disabled`

Best Regards,
Sacha Bertschi

mvanhorn:
## Summary

Adds support for the `TZDATA_AUTOUPDATE` environment variable to disable tzdata's automatic update checks. Set `TZDATA_AUTOUPDATE=disabled` (or `false` or `0`) to prevent the autoupdate.

## Why

In read-only container environments like RedHat OpenShift, tzdata's autoupdate attempts to write to an internal file and fails repeatedly, flooding pod logs with errors every few seconds.

## Changes

`config/runtime.exs`: Added a check before the test config block that reads `TZDATA_AUTOUPDATE` and configures `:tzdata, :autoupdate, :disabled` when set. Runs in all environments (not just prod) since containers can run in any Mix env.

## Usage

```bash
TZDATA_AUTOUPDATE=disabled mix phx.server
```

Fixes #519

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -340,6 +340,10 @@ if config_env() == :prod do
   end
 end
 
+if System.get_env("TZDATA_AUTOUPDATE") in ["disabled", "false", "0"] do
+  config :tzdata, :autoupdate, :disabled
+end
+
 if config_env() == :test do
   db_url = System.get_env("DB_URL")
 


wmnnd:
Hey, thank you for creating the PR.

As I wrote in the issue, the env variable should be called `DISABLE_TZDATA_UPDATES` and you should use the same convention for determining whether it's set as the other `DISABLE_...` variables.
Also, please sign the CLA as described here: https://github.com/pentacent/keila/blob/main/.cla/README.md

wmnnd:
Uhm, actually, why did you open this PR when there is already #520? I will close this in favor of the other PR.

PR #526, opened 2026-04-20
Added a prefix (postgresql schema) from env variable into runtime.exs…

sachabertschibfs:
… and also directly into the Oban migration.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -90,7 +90,10 @@ config :logger, :console,
 # Use Jason for JSON parsing in Phoenix
 config :phoenix, :json_library, Jason
 
+db_schema = System.get_env("DB_SCHEMA")
+prefix = if db_schema in [nil, ""], do: "public", else: db_schema
 config :keila, Oban,
+  prefix: prefix,
   queues: [
     mailer: 50,
     mailer_scheduler: 1,
--- a/config/config.exs
+++ b/config/config.exs
@@ -90,10 +90,7 @@ config :logger, :console,
 # Use Jason for JSON parsing in Phoenix
 config :phoenix, :json_library, Jason
 
-db_schema = System.get_env("DB_SCHEMA")
-prefix = if db_schema in [nil, ""], do: "public", else: db_schema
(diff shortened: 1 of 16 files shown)


wmnnd:
Hey there, thanks for the PR! There are actually multiple Oban migrations (try looking for `Oban.Migrations.up`). Do they work without the prefix?

sachabertschibfs:
Oops, I missed those, I'll add them.

sachabertschibfs:
I found a few more prefix to deal with. It should now migrate without problem and run.
It would be best if you test it yourself of course ^^

sachabertschibfs:
Hello @wmnnd ,
Could you review the pr?
I believe the migrations and the application run fine with a different schema than public now.
Best Regards,
Sacha Bertschi

wmnnd:
Hey there, the current implementation doesn't work because you've moved the schema settings to config.exs. However, config.exs is run only at compile-time so it won't be possible to change this at runtime (i.e. when starting as a container).

wmnnd:
I think the following code in runtime.exs should be sufficient (config merges keys automatically):

```elixir
db_schema = System.get_env("DB_SCHEMA")
if db_schema not in [nil, ""] do
  config :keila, Keila.Repo,
    migration_default_prefix: db_schema,
    parameters: [search_path: db_schema]

  config :keila, Oban, prefix: db_schema
end
```

I suggest it you put it underneath the existing `# Database` try/catch block and add a `#Database prefix` comment,

Could you give that a try?

sachabertschibfs:
Hello @wmnnd ,
I changed to put the configuration in the runtime as you said.
I had tried different things, I ended putting it in config which was wrong as you stated.
Should be good now.

Best Regards,
Sacha Bertschi

sachabertschibfs:
Hello @wmnnd ,
Had you the time to review the changes ?

Best Regards,
Sacha Bertschi

wmnnd:
It's looking good, so I'll merge it now. Let me know if this works in your environment.

PR #530, opened 2026-05-06
Add pt br translation

polesapart:
Add pt-BR translation. Signed CLA just in case it's needed.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -123,7 +123,7 @@ config :mime, :types, %{
 # Configure locales
 config :keila, KeilaWeb.Gettext,
   default_locale: "en",
-  locales: ["de", "en", "es", "fr", "hu", "it", "bg"]
+  locales: ["de", "en", "es", "fr", "hu", "it", "bg", "pt_BR"]
 
 config :ex_cldr,
   default_backend: Keila.Cldr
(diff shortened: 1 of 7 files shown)


wmnnd:
Olá @polesapart, thank you for contributing this translation! Portuguese would be a fantastic addition to Keila! 🇧🇷🇵🇹

I had an LLM look over it and it only had two notes:
```
admin.po: "Login as" → "Logar como" — should be "Fazer login como"
auth.po: "Log in with your new password" → "Logar com sua nova senha" — should be "Entrar com sua nova senha" or "Fazer login com sua nova senha"
```
Since I don't speak Portuguese, I will leave it up to you whether this suggestion makes sense or whether your original translation is better. Just let me know ☺️

wmnnd:
Oh, and one more thing: Your PR touches a bunch of other translation files that are not directly connected to your translation. Could you please reset them again so that this branch only includes the changes necessary to add your translation?

polesapart:
Hi!

The translation part is fine, both meanings are correct, but the shorter form is usually preferably, specially if used on buttons or other space constrained gui elements.

I'll look at the other files you mentioned once I'm in front of pc.

Glad I can help!

polesapart:
I think I managed to rollback the affected, unrelated files now!

PR #549, opened 2026-07-25
Add option to generate embedded form markup without styles

### LINKED ISSUE #114: Add option to generate embedded form markup without styles
There should be a checkbox in the form generator for generating form markup without styles.

In a later iteration, this could be expanded to let users choose between the following options:
- no styles
- with tailwind styles
- with inline styles only

JoeJoeflyn:
This PR adds a `styles_disabled` option to form settings. When enabled, the embed code is generated without Tailwind classes and inline styles, so it inherits the host site's CSS.

The hosted form page and editor preview are unaffected, only the embed code changes.

fixes #114


wmnnd:
Thank you for creating this PR!

As mentioned in the #114, I would like to keep the option open to add inline styles later. So I think using an enum instead of a boolean would be better. We could use `embedded_styles`: `enabled` (default) | `disabled` | `inline` for that.
My suggestion for the helper function that you currently have (maybe_class/2) would to be refactor it into `styles/2` where the first argument is `form` and the second is the classes. Once we add support for inline styles, a third argument would be the inline styles. The helper function could pattern-match on the form directly so that calling it with a form that has embedded_styles set to `disabled` returns `[]` while the default clause returns `[class: class]`. No need to add the inline option just yet.

JoeJoeflyn:
Thanks! Updated as suggested, enum + `styles/2` helper, ready for another look when you have a moment.

lib/keila/contacts/schemas/form_settings.ex
```diff
@@ -30,6 +30,7 @@ defmodule Keila.Contacts.Form.Settings do
     field(:welcome_enabled, :boolean, default: false)
     field(:welcome_subject, :string)
     field(:welcome_markdown_body, :string)
+    field(:embedded_styles, :string, default: "enabled")
```
wmnnd: This should be Ecto.Enum instead of `:string`.

lib/keila_web/templates/form/edit_live.html.heex
```diff
@@ -154,6 +154,20 @@
             {gettext("Allow embedding as HTML")}
           <% end %>
         <% end %>
+        <%= with_validation(fs, :embedded_styles) do %>
```
wmnnd: I think it would make more sense to put this select underneath the textarea where the embed code is printed - and only show it if embedding is allowed (just like the textarea itself).

lib/keila_web/templates/form/edit_live.html.heex
```diff
@@ -154,6 +154,20 @@
...
+          <%= label(fs, :embedded_styles) do %>
+            {gettext("Embedded form styles")}
+          <% end %>
+          {select(
+            fs,
+            :embedded_styles,
+            [
+              {"enabled", gettext("Enabled (Tailwind classes + inline styles)")},
```
wmnnd: The label should just read: "Include Tailwind classes"/"Disable Tailwind classes".

lib/keila_web/views/public_form_view.ex
```diff
@@ -83,20 +81,32 @@ defmodule KeilaWeb.PublicFormView do
...
 
-  defp build_form_styles(form) do
-    build_styles(%{
-      "background-color" => form.settings.form_bg_color,
-      "color" => form.settings.text_color
-    })
-  end
+  defp build_form_styles(%{settings: %{embedded_styles: "disabled"}}), do: nil
+
+  defp build_form_styles(form),
```
wmnnd: This should not use the `do:` shorthand because it's multiple lines. It should just stay as it was before.

lib/keila_web/views/public_form_view.ex
```diff
@@ -83,20 +81,32 @@ defmodule KeilaWeb.PublicFormView do
     end
   end
 
-  defp build_form_styles(form) do
-    build_styles(%{
-      "background-color" => form.settings.form_bg_color,
-      "color" => form.settings.text_color
-    })
-  end
+  defp build_form_styles(%{settings: %{embedded_styles: "disabled"}}), do: nil
```
wmnnd: This should return `[]`. Like that you don't have to do the whole `maybe_put_opt` thing and can keep the code for setting styles as it was. Besides, I the way I think about the feature, it's only about disabling Tailwind classes - the inline styles that have been configured should stay in place. Do you think it makes more sense to disable these inline styles as well when we're removing the Tailwind classes?

wmnnd (changes requested):
Thanks! I've left a few more notes. The one option question is whether disabling Tailwind styles really should remove all styles or not.

JoeJoeflyn:
Thanks! All addressed.

lib/keila_web/views/public_form_view.ex
JoeJoeflyn: I agreed, only Tailwind classes are removed, inline styles stay

PR #552, opened 2026-07-27
Fix deleting contacts by external_id via API

### LINKED ISSUE #548: Deleting a contact via API using the external_id
Deleting a contact via API using the `external_id` to identify it does not work, contrary to what is indicated in the API documentation.

The issue is that deleting a contact via API uses the standard contact deletion function defined in `contacts.ex`. 
The delete API call sets the filter to `external_id`. 
However, in `query.ex`, `external_id` is not included among the possible fields for filters.

I tried adding it to `query.ex`, and that solved the problem:

```elixir
@fields ["id", "email", "inserted_at", "first_name", "last_name", "status", "double_opt_in_at", "external_id"]
```

If you think the solution is correct I can open a PR

JoeJoeflyn:
`external_id` was missing from the `@fields` list in `Keila.Contacts.Query`, so filtering by `external_id` (which the delete API endpoint does when `id_type=external_id`) silently matched nothing and the contact was not deleted.

Adds `external_id` to `@fields` so the filter builds a valid query.

Closes #548.


wmnnd:
Thanks for finding this bug and fixing it! I saw you signed the CLA in the other PR, so I'll merge this after that one has been merged.

JoeJoeflyn:
tks, appreciate it

PR #553, opened 2026-07-27
Fix update check crashing on v-prefixed release tags

### LINKED ISSUE #546: Update check silently stops working after tag names got a "v" prefix (breaks since v0.19.1)
**Summary**

  The built-in "you are running the latest version" check (Keila.Instance.fetch_updates/0) stops working as soon as it encounters a GitHub release whose tag_name has a leading v — which is every release since v0.19.1 (2026-04-15). Instances that were already running before that release will silently keep reporting "up to date" forever afterwards, even though several newer versions (v0.20.0, v0.20.1, v0.30.0, v0.30.1, v0.30.2, ...) have since been published.

**Probable cause**
  
lib/keila/instance/schemas/release.ex validates each release's version string with Elixir's Version.parse/1:

  validate_change(:version, fn :version, version ->
    case Version.parse(version) do
      {:ok, _} -> []
      _ -> [version: "invalid version format"]
    end
  end)

Version.parse/1 requires strict semver (0.30.2) and rejects a v-prefixed string like v0.30.2. Since tags moved to the v-prefixed format starting with v0.19.1, the changeset built in Release.new!/1 is invalid for every current and future release.

Release.new!/1 calls Ecto.Changeset.apply_action!(:insert), which raises Ecto.InvalidChangesetError on an invalid changeset. That exception is not caught anywhere in Keila.Instance.fetch_updates/0's with pipeline (the else branch there only handles pattern-match failures, not raised exceptions), so the whole function crashes.

Since fetch_updates/0 is invoked from Keila.Instance.UpdateCronWorker (an Oban job run daily via Oban.Plugins.Cron at "1 0 * * *", plus once at boot), the crash means available_updates on the Instance record never gets written past whatever value it last held from before the v prefix was introduced. The admin UI then keeps showing the old (often empty) list indefinitely, i.e. it reports "running latest version" even when it isn't.

**Steps to reproduce (confirmed live)**

Ran directly against a live instance via bin/keila rpc "Keila.Instance.fetch_updates()", hitting the real GitHub API:

  ** (Ecto.InvalidChangesetError) could not perform insert because changeset is invalid.

  Changeset: ... errors: [version: {"invalid version format", []}] ...
  data: #Keila.Instance.Release<>, valid?: false

      (ecto 3.13.5) lib/ecto/changeset.ex:2323: Ecto.Changeset.apply_action!/2
      (elixir 1.18.4) lib/enum.ex:1714: Enum."-map/2-lists^map/1-1-"/2
      (keila 0.30.2) lib/keila/instance.ex:65: Keila.Instance.fetch_updates/0

  This confirms the crash happens on real release data (v0.30.2, the current latest tag) — not just a theoretical read of the code. Also confirmed directly:
  Version.parse("v0.30.2") → :error, Version.parse("0.30.2") → {:ok, ...}.

**Expected behaviour**

  The update check should recognise v-prefixed tags (e.g. by stripping a leading v before calling Version.parse/1) and correctly report the release as newer when it is. It should also fail gracefully (log + leave available_updates untouched, rather than crash the whole job) if any single release entry can't be parsed, so one bad/unexpected tag format doesn't take down the entire check.

**Suggested fix**

  In lib/keila/instance/schemas/release.ex, normalize the version string before parsing, e.g.:

  defp normalize_version(v), do: String.trim_leading(v, "v")

  and apply it both in the validate_change callback and wherever version is cast/stored — plus ideally wrap the Enum.map(&Release.new!/1) step in Keila.Instance.fetch_updates/0 with a rescue/filter so a single malformed release entry (from either the GitHub API or a future tagging convention change) doesn't crash the whole check.

**Environment**
  
  - Reproduced on Keila 0.30.2 itself (the crash isn't limited to old versions — it happens on the current release checking itself against its own latest tag).
  - Confirmed via GitHub API: v0.30.2, v0.30.1, v0.30.0, v0.20.1, v0.20.0, v0.19.1 all have the v prefix; 0.19.0 and earlier do not.
  - Deployed via the official pentacent/keila Docker image, docker compose.

JoeJoeflyn:
Since v0.19.1, Keila's git tags use a `v` prefix (`v0.30.2` instead of `0.30.2`). Elixir's `Version.parse/1` requires strict semver and rejects the `v` prefix, so `Release.changeset` marked every current release as invalid and `Release.new!/1` raised `Ecto.InvalidChangesetError`. The exception was uncaught in `fetch_updates/0`, silently breaking the daily update check on every deployed instance — the admin UI keeps reporting "up to date" even when it isn't.

Strips a leading `v` from the version string in `Release.changeset` before validation, and switches `fetch_updates/0` to the non-raising `Release.new/1` so a single malformed release is logged and skipped instead of crashing the whole job.

Closes #546.

### DIFF
--- a/lib/keila/instance.ex
+++ b/lib/keila/instance.ex
@@ -62,7 +62,16 @@ defmodule Keila.Instance do
       |> Enum.map(fn %{"tag_name" => version, "published_at" => published_at, "body" => changelog} ->
         %{version: version, published_at: published_at, changelog: changelog}
       end)
-      |> Enum.map(&Release.new!/1)
+      |> Enum.flat_map(fn attrs ->
+        case Release.new(attrs) do
+          {:ok, release} ->
+            [release]
+
+          {:error, changeset} ->
+            Logger.info("Unable to parse release: #{inspect(changeset.errors)}")
+            []
+        end
+      end)
       |> Enum.filter(fn %{version: version} ->
         version |> Version.parse!() |> Version.compare(current_version) == :gt
       end)
--- a/lib/keila/instance/schemas/release.ex
+++ b/lib/keila/instance/schemas/release.ex
@@ -10,6 +10,7 @@ defmodule Keila.Instance.Release do
   def changeset(struct \\ %__MODULE__{}, params) do
     struct
     |> cast(params, [:version, :published_at, :changelog])
+    |> update_change(:version, &normalize_version/1)
     |> validate_required([:version, :published_at, :changelog])
     |> validate_change(:version, fn :version, version ->
       case Version.parse(version) do
@@ -22,4 +23,14 @@ defmodule Keila.Instance.Release do
   def new!(params) do
     params |> changeset() |> Ecto.Changeset.apply_action!(:insert)
   end
+
+  def new(params) do
+    params |> changeset() |> Ecto.Changeset.apply_action(:insert)
+  end
+
+  defp normalize_version(version) when is_binary(version) do
+    String.trim_leading(version, "v")
+  end
+
+  defp normalize_version(version), do: version
 end
--- a/test/keila/instance/release_test.exs
+++ b/test/keila/instance/release_test.exs
@@ -1,0 +1,53 @@
+defmodule Keila.Instance.ReleaseTest do
+  use ExUnit.Case, async: true
+
+  alias Keila.Instance.Release
+
+  describe "new!/1" do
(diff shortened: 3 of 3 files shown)


wmnnd:
Thank you for contributing this PR! It's actually a decent solution - but fortunately I was able to simply modify the existing releases; and a one-line change in the workflow file should fix this for future releases. That's why I'll close this PR now without merging it 😊

PR #561, opened 2026-08-20
fix: Disable STARTTLS for the system mailer when configured off

### LINKED ISSUE #536: Bug: MAILER_ENABLE_STARTTLS=false does not actually disable STARTTLS
### Problem

When `MAILER_ENABLE_STARTTLS` is set to `false`, STARTTLS is not actually disabled. The underlying `gen_smtp` library will attempt STARTTLS anyway if the SMTP server advertises it in the EHLO response.

This happens because in `config/runtime.exs`, the `else` branch of the `starttls?` check simply returns the config unchanged, without explicitly setting `:tls` to `:never`:

```elixir
|> then(fn config ->
  if starttls? do
    config
    |> Keyword.put(:tls, :always)
    ...
  else
    config  # tls option is not set, gen_smtp defaults to attempting STARTTLS if offered
  end
end)
```

When `gen_smtp` does not find a `:tls` option in the config, it falls back to its default behavior, which is to attempt STARTTLS if the server offers it. This causes connection failures when TLS negotiation fails, even though the user explicitly set `MAILER_ENABLE_STARTTLS=false`.

### Fix

Explicitly set `:tls` to `:never` in the `else` branch:

```elixir
|> then(fn config ->
  if starttls? do
    config
    |> Keyword.put(:tls, :always)
    |> Keyword.put(
      :tls_options,
      :tls_certificate_check.options(host) ++ [versions: [:"tlsv1.2"]]
    )
  else
    config
    |> Keyword.put(:tls, :never)
  end
end)
```

### How to reproduce

1. Set `MAILER_ENABLE_STARTTLS=false` and `MAILER_ENABLE_SSL=false` in your environment
2. Use an SMTP server (e.g. Postfix) that advertises STARTTLS in its EHLO response
3. Observe that Keila attempts STARTTLS anyway, resulting in a connection error

### Notes

This issue is particularly relevant for self-hosted setups where Keila connects to a local SMTP server for system emails (e.g. Postfix with `inet_interfaces = localhost`) and TLS is not needed or desired for local connections.

*********

If you think my change is okay, I can start a PR. 
But I have a question regarding the CLA.
My GitHub account belongs to an organization account, my company. 
In the CLA signature, can I use my username as legal name, or would you prefer the businessname of the company I work for?

mvanhorn:
Change the false branch of the system-mailer STARTTLS configuration in `config/runtime.exs` to add `tls: :never`, while leaving the existing `tls: :always` and certificate options for enabled STARTTLS untouched. The system SMTP mailer reads `MAILER_ENABLE_STARTTLS`, but its false branch leaves the Swoosh/gen_smtp `:tls` option unset.

With `MAILER_ENABLE_STARTTLS=false` and `MAILER_ENABLE_SSL=false`, the resolved `Keila.Auth.Emails` SMTP config contains `tls: :never` and does not contain STARTTLS certificate options; With `MAILER_ENABLE_STARTTLS=true`, the resolved SMTP config still contains `tls: :always` and the existing TLS certificate/version options.

Resolves #536

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -121,7 +121,7 @@ if config_env() == :prod do
                 :tls_certificate_check.options(host) ++ [versions: [:"tlsv1.2"]]
               )
             else
-              config
+              Keyword.put(config, :tls, :never)
             end
           end)
       end
--- a/test/config/runtime_test.exs
+++ b/test/config/runtime_test.exs
@@ -1,0 +1,95 @@
+defmodule Keila.Config.RuntimeTest do
+  use ExUnit.Case, async: true
+
+  @smtp_host "smtp.example.test"
+
+  test "disables STARTTLS when MAILER_ENABLE_STARTTLS is false" do
+    config = runtime_mailer_config(MAILER_ENABLE_STARTTLS: "false", MAILER_ENABLE_SSL: "false")
+
+    assert config[:tls] == :never
+    refute Keyword.has_key?(config, :tls_options)
+  end
+
+  test "configures STARTTLS certificate options when MAILER_ENABLE_STARTTLS is true" do
+    config = runtime_mailer_config(MAILER_ENABLE_STARTTLS: "true", MAILER_ENABLE_SSL: "false")
+
+    assert config[:tls] == :always
+
+    assert config[:tls_options] ==
+             [certificate_options_match?: true, versions: [:"tlsv1.2"]]
+  end
+
+  test "retains direct SSL options when STARTTLS is disabled" do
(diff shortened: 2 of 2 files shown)


wmnnd:
Thank you for giving this a shot! I have thought about this some more and decided to go a slightly different route, ensuring backwards compatibility. In 7bbb46dff2a3ac3af3117ebab173ff81732001ba, I have thus introduced two new env vars: `MAILER_SMTP_TLS_MODE` and `MAILER_SMTP_AUTH_METHOD`. If you set `MAILER_SMTP_TLS_MODE` to `none`, Keila will not attempt to use STARTTLS.

